PR #5, opened 2017-05-04
Tsw 62495

fwang-teradici:


dev/domain-controller/FinalizeDeployment.ps1
```diff
@@ -1,10 +1,21 @@
 ﻿Param(
+  [string]$rgName,
   [string]$domainUserUPN,
   [string]$domainPassword,
   [string]$machineToJoin,
-  [string]$groupToJoin
+  [string]$groupToJoin,
+  [string]$azureUserName, 
+  [string]$azurePassword,
```
bill-dall: Can't we pass these as a credential or securestring?

dev/domain-controller/FinalizeDeployment.ps1
```diff
@@ -1,10 +1,21 @@
 ﻿Param(
+  [string]$rgName,
   [string]$domainUserUPN,
   [string]$domainPassword,
```
bill-dall: I guess the same comment goes for what I did here :)

dev/domain-controller/FinalizeDeployment.ps1
```diff
@@ -20,3 +31,39 @@ Invoke-Command -Session $psSession -ScriptBlock {
...
+$parameters.Add(“subnetRef”, $subnetRef)
+$parameters.Add(“skuName”, "Standard_Small")
+$parameters.Add(“capacity”, 1)
+$parameters.Add(“backendIpAddressDefault”, "$backendIpAddressDefault")
+$parameters.Add(“backendIpAddressForPathRule1”, "$backendIpAddressForPathRule1")
+$parameters.Add(“pathMatch1”, "/pcoip-broker/*")
+$parameters.Add(“certData”, "$fileContentEncoded")
+$parameters.Add(“certPassword”, "$randomPswd")
+
+$deployName = -join ((65..90) + (97..122) | Get-Random -Count 12 | % {[char]$_})
```
bill-dall: So I think we can call this 'deploy-app-gateway' instead of a random thing? That should work since deployment names are local to the resource group.

fwang-teradici: it normally work unless the parent deploy name is the same. I choose the random name to prevent the name conflict with parent name and this name does not show in portal

dev/domain-controller/FinalizeDeployment.ps1
fwang-teradici: I followed the parameter domainUserUPN and domainPassword defined. the commandToExecute which will call this ps file  concats command and all parameters together, so not sure if we can use credential

bill-dall:
Hmm- I'm worried about the changes in azuredeploy.json. There are a few other outstanding pull requests so the re-formatting is going to cause some conflicts. But probably most important is that I think it's making a keyvault we no longer need?

dev/domain-controller/FinalizeDeployment.ps1
bill-dall: Oh right- I'd make the whole 'commandToExecute' a protected setting. Duh. We're okay. Thanks for the reminder.

fwang-teradici:
the only one keyvault created is at Install-CAM, I am not sure how to use that keyvault. I did not create keyvault anymore at this moment until we want to keep the generated certificate.

PR #16, opened 2017-08-03
Tsw 64738

fwang-teradici:


dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -303,13 +313,13 @@ Configuration InstallCAM
...
 				{
-				  $NewPath= $using:JavaBinLocation + �;� + $NewPath
+				  $NewPath= $using:JavaBinLocation + ";" + $NewPath
 				}
 
 				#these get added to the environment on next reboot
-				Set-ItemProperty -Path "$Reg" -Name PATH �Value $NewPath
-				Set-ItemProperty -Path "$Reg" -Name JAVA_HOME �Value $using:JavaRootLocation
-				Set-ItemProperty -Path "$Reg" -Name classpath �Value $using:JavaLibLocation
+				Set-ItemProperty -Path "$Reg" -Name PATH -Value $NewPath
```
bill-dall: This is just changing from one '-' to another type of '-'?

dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -1320,7 +1330,159 @@ brokerLocale=en_US
...
+	using System.Security.Cryptography.X509Certificates;
+	public class TrustAllCertsPolicy : ICertificatePolicy {
+		public bool CheckValidationResult(
+			ServicePoint srvPoint, X509Certificate certificate,
+			WebRequest request, int certificateProblem) {
+			return true;
+		}
+	}
+"@
+				[System.Net.ServicePointManager]::CertificatePolicy = New-Object TrustAllCertsPolicy
```
bill-dall: Which context does this get changed for? Just the current executing environment? PS Session?

dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -1320,7 +1330,159 @@ brokerLocale=en_US
...
+				##
+
+				# Read in Authorization Information
+				Get-Content "$env:AZURE_AUTH_LOCATION" | Foreach-Object{
+					$var = $_.Split('=', 2)
+					New-Variable -Name $var[0] -Value $var[1]
```
bill-dall: Would be good to say which variables you are expecting to be populated here.

dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -1320,7 +1330,159 @@ brokerLocale=en_US
...
+						} catch {
+							$registerUserResult = ConvertFrom-Json $_.ErrorDetails.Message
+						}
+						Write-Verbose (ConvertTo-Json $registerUserResult)
+						# Check if registration succeeded or if it has been registered previously
+						if( !(($registerUserResult.code -eq 201) -or ($registerUserResult.data.reason.ToLower().Contains("already exist"))) ) {
+							throw ("Failed to register with CAM. Result was: " + (ConvertTo-Json $registerUserResult))
+						}
+
+						[System.Environment]::SetEnvironmentVariable("CAM_USERNAME", $userRequest.username, "Machine")
```
bill-dall: Much nicer than the registry setting I was doing. So if you set 'machine' it doesn't change for the running process but will change for any new processes (and after reboot, since it's setting the registry)?

dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -1320,7 +1330,159 @@ brokerLocale=en_US
...
+						if ($signInResult.code -ne 200) {
+							throw ("Signing in failed. Result was: " + (ConvertTo-Json $signInResult))
+						}
+						$tokenHeader = @{
+							authorization=$signInResult.data.token
+						}
+						Write-Host "Cloud Access Manager sign in succeeded"
+
+						$rcCred = $using:registrationCodeAsCred
+						$registrationCode = $rcCred.Password
```
bill-dall: Does that work? I thought you needed to 'getnetworkcredentials.password' or something like that otherwise the blob stays encrypted.

dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -1320,7 +1330,159 @@ brokerLocale=en_US
...
+								$registeredDeployment = ConvertFrom-Json $_.ErrorDetails.Message
+								throw ("Getting Deployment ID failed. Result was: " + (ConvertTo-Json $registeredDeployment))
+							}
+						} else {
+							$deploymentId = $registerDeploymentResult.data.deploymentId
+						}
+						[System.Environment]::SetEnvironmentVariable("CAM_DEPLOYMENTID", $deploymentId, "Machine")
+						Write-Host "Deployment has been registered succesfully with Cloud Access Manager"
+
+						# Register Agent Machine
```
bill-dall: I think that machine is being made in parallel - any race conditions here?

dev/domain-controller/new-admin-vm/Install-CAM.ps1
cptera: it looks like guithub and linux don't like whatever encoding primesoft is using for the files so the get rendered weird in the brower and on my kubuntu machine. This is just to "correct" it so it looks and runs right. So yes, '-' to '-'

dev/domain-controller/new-admin-vm/Install-CAM.ps1
cptera: yes, just the current PS session

dev/domain-controller/new-admin-vm/Install-CAM.ps1
bill-dall: Primesoft hasn't been modifying these files - so me probably. Thanks for cleaning up.

fwang-teradici:
Hi,

I do not think they use ISO8859-1 or UTF-8 encode. They might use different encode

Thanks
Frank

dev/domain-controller/new-admin-vm/Install-CAM.ps1
cptera: sure

dev/domain-controller/new-admin-vm/Install-CAM.ps1
cptera: yes

dev/domain-controller/new-admin-vm/Install-CAM.ps1
cptera: you're right, it was passing the object type, not the registration code. will fix.

cptera: I copied this from line 877 so I'm not sure if there's a bug there or not?

dev/domain-controller/new-admin-vm/Install-CAM.ps1
fwang-teradici: I see the password save in environment which is a security issue

cptera: @fwang-teradici You're right - i think we're accepting it though because we need to provide the password to the java applications somehow

fwang-teradici: All register user, deployment and machine in the same retry for loop which means each one fail will cause retry other already successful operation. Also to test if existing case shall use code instead of string comparing

cptera: all in the same loop should be fine - it will return "balh blah blah alread exists" and we continue from there. It's a bit lazy, but it was quicker and cleaner than wrapping a retry around every Rest call. 
I'm not sure if checking the code is sufficient since the same code could be caused by multiple things - I feel more comfortable using the error reason.

cptera:
@bill-dall , @fwang-teradici could you make any comments about the antar branch in opull request 15 instead of here? this pull request was supposed to be just for frank changing the agent version and I've merged that into the antar branch already

bill-dall:
Okay.

dev/domain-controller/new-admin-vm/Install-CAM.ps1
fwang-teradici: 409 is only used for duplicated case

PR #17, opened 2017-08-16
Added auto-shutdown capabilities (disabled for now)

cptera:


dev/domain-controller/new-agent-vm/server2016-graphics-agent.json
```diff
@@ -240,7 +241,8 @@
                 "videoDriverUrl": "[variables('videoDriverUrl')]",
                 "gitLocation": "[parameters('_artifactsLocation')]",
                 "sumoCollectorID": "[concat('CAMPOC-',resourceGroup().name,'-Agent-',subscription().subscriptionId)]",
-                "domainGroupToJoin": "[parameters('domainGroupToJoin')]"
+                "domainGroupToJoin": "[parameters('domainGroupToJoin')]",
+                "enableAutoShutdown": "[variables('enableAutoShutdown')]"
```
bill-dall: Ahh - So this default to empty string is 'disabled.'

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.ps1
```diff
@@ -372,6 +375,74 @@ Configuration InstallPCoIPAgent
...
+
+            TestScript = { 
+				$serviceName = "CAMIdleShutdown"
+				
+				if ([string]::IsNullOrWhiteSpace($using:enableAutoShutdown)) {
+					return $true
+				}
+
+				try {
+					$svc = Get-WmiObject -Class Win32_Service -Filter "Name='$ServiceName'"
```
bill-dall: Do we need an -erroraction stop inside the try block here? Traditionally I have seen that as needed for exception handling to work when DSC is invoked (which is different from standard script execution subtly).

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.ps1
```diff
@@ -372,6 +375,74 @@ Configuration InstallPCoIPAgent
...
+					throw $msg
+				}
+
+				$svc = Get-Service -Name $serviceName   
+
+				if ($svc.StartType -ne "Automatic") {
+					$msg = "try setting {0} Service start type to automatic." -f $serviceName
+					Write-Verbose $msg
+
+					Set-Service -name  $serviceName -StartupType Automatic
```
bill-dall: Similar comment to above -if 'Set-Service' has an error then even if you have error handling code after, DSC seems to sense the error (a disturbance in the force) and bork. This is a real PITA to test, naturally - did you see this handle these error conditions properly when you did your testing?

dev/domain-controller/new-agent-vm/server2016-graphics-agent.json
fwang-teradici: It seems to use bool type and set default value to false is better choise

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.ps1
cptera: You're right - need to add the error action for the case where the service isn't installed

dev/domain-controller/new-agent-vm/server2016-graphics-agent.json
cptera: done

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.ps1
cptera: I copied most of the Service setup stuff from another Script from the DSC. I haven't seen any errors related to this and we already try to get the service and check that it exists (by checking the StatType) so I don't think we need any additional error handling

PR #18, opened 2017-08-17
TSW-65735 deploy fail

bill-dall:


dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -1089,8 +1070,8 @@ graphURL=https\://graph.windows.net/
 
 				$catalinaHome = $using:CatalinaHomeLocation
 				$catalinaBase = "$catalinaHome\$using:brokerServiceName"
-				# I don't think we need to set the registry
-				# Set-ItemProperty -Path "$Reg" -Name CATALINA_BASE -Value $using:CatalinaHomeLocation
+
+                #set the current (temporary) environment
```
fwang-teradici: is it necessary to reset here because it is set in Install_Tomcat and this script depends on Install_Tomcat

bill-dall: CATALINA_BASE? That's no longer set in Install_Tomcat since the two instances have different CATALINA_BASE's

dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -394,23 +387,15 @@ Configuration InstallCAM
...
-				  $NewPath= $NewPath + ";" + $CatalinaBinLocation
+				    #put tomcat path at the end
+				    $NewPath= $NewPath + ";" + $CatalinaBinLocation
 				}
 
-				Set-ItemProperty -Path "$Reg" -Name PATH -Value $NewPath
-				Set-ItemProperty -Path "$Reg" -Name CATALINA_HOME -Value $CatalinaHomeLocation
-
-				# set the current environment CATALINA_HOME as well since the service installer will need that
-				$env:CATALINA_HOME = $CatalinaHomeLocation
```
bill-dall: Actually removing this line was a BAD idea since [System.Environment]::SetEnvironmentVariable(..."Machine") doesn't actually set the local process environment. :( I have a change I'm testing now.

dev/domain-controller/new-admin-vm/Install-CAM.ps1
fwang-teradici: then that is OK

PR #20, opened 2017-08-25
Tsw65834 deployment failures

bill-dall:


dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -529,10 +526,20 @@ Configuration InstallCAM
 				$CatalinaHomeLocation = $using:CatalinaHomeLocation
 				$catalinaBase = "$CatalinaHomeLocation" #\$using:AUIServiceName"
 
-                Write-Verbose "Install Nuget and AzureRM packages"
+                Write-Verbose "Ensure Nuget Package Provider and AzureRM module are installed"
 
-				Install-packageProvider -Name NuGet -Force
-				Install-Module -Name AzureRM -Force
+				If(-not [bool](Get-PackageProvider -ListAvailable | where {$_.Name -eq "NuGet"}))
```
bill-dall: I got the 'no AzureRM module' error again. Clearly there is a module and it seems to already be installed so the most conservative option seemed to be to just pull the package provider and module if needed. We'll have to see if it fixes things. The best reason I can come up with for the intermittent failure is that NuGet is still getting configured when the InstallModule is called so it fails strangely.

dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -1329,15 +1331,13 @@ brokerLocale=en_US
 				& "keytool" -import -file "$env:systemdrive\$issuerCertFileName" -keystore ($env:classpath + "\security\cacerts") -storepass changeit -noprompt
                 $ErrorActionPreference = $eap
 
-		        Write-Host "Finished! Restarting Tomcat service for $using:brokerServiceName."
-
-				Restart-Service $using:brokerServiceName
```
bill-dall: Not starting the services (ever) since the final step will reboot the machine and start the services that way. They're both set to Automatic.

PR #21, opened 2017-08-28
Tsw 65526

fwang-teradici:
change CAM SAAS URI to use parameter and default point to staging https://cam-staging.teradici.com


fwang-teradici:
I believe this is for antar

PR #22, opened 2017-09-07
TSW-65526 add parameter verifyCAMSaaSCertificate with default value true

fwang-teradici:


dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -1514,6 +1521,8 @@ brokerLocale=en_US
 					throw $camRegistrationError
 				}
 
+				# restore CertificatePolicy
```
bill-dall: Why restore? What is the longevity and scope of the change? Note also that the Java for AUI and Broker are also consuming CAM SaaS API's so need to get an environment value with the setting.

fwang-teradici: once done register with CAM, it has to restore to the original certificate policy before making changes, otherwise it might leave security hole on the machine in windows powershell environment. The scope is limit to register CAM Front End with SaaS. This is not related with Java for AUI and Broker. If Broker and Java for AUI needs to skip verifying SaaS server certificate when consuming SaaS API, both needs to be configured individually

bill-dall: I'll make a backlog story for the Broker and AUI.

PR #26, opened 2017-09-08
* Fixed error in DSC where it doesn't raise the correct error string

cptera:


cptera:
Yes, this fixed the problem. Eg: https://cloudcity.teradici.com/viewLog.html?buildId=5011&buildTypeId=CamFrontEnd_DeployCAMFEServicePrincipalAccount&tab=buildLog&_focus=469 returned the error correctly - previously it would say "Input String is in wrong format" as seen in https://cloudcity.teradici.com/viewLog.html?buildId=4970&buildTypeId=CamFrontEnd_DeployCAMFEServicePrincipalAccount&tab=buildLog

bill-dall:
Thanks! Weird. Okay.

PR #28, opened 2017-09-11
TSW-66322 – added more informative error messages

bill-dall:
TSW-66322 – added more informative error messages
TSW-66317 – fix test-script issues around finding files and folders

After we merge all of these I guess for now only Frank can create the .ps1.zip files?


bill-dall:
Good that should help between all the changes. Now what are we going to use for a binaries staging environment for smoke testing these?

PR #29, opened 2017-09-12
Tsw 66292 split arm deployment template

bill-dall:
Hi guys,

Since I've had a couple deployments working in a row now I thought I'd start the pull request to show you what I've been up to.

As I mentioned, some things are hard-coded variables that need to be parameters. That can/will be fixed later. The main thing here is to reduce 'big branches' and have working code with the same performance.

This does change the URL of the azuredeploy.json. So before merging or concurrently we'd want to update the TeamCity tests.

If you approve this pull request I won't merge until we've discussed at a standup, so feel free to review/approve/request changes as normal.


bill-dall:
I'll cancel this one as #32 supersedes it.

PR #31, opened 2017-09-28
Tsw 65310 deploy sa for linux

jackyycheng:


dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -193,6 +196,12 @@ Configuration InstallCAM
 			DestinationPath = "$LocalDLPath\$gaAgentARM"
 		}
 
+		xRemoteFile Download_Linux_Agent_ARM
+		{
+			Uri = "$templateAgentURI/$linuxAgentARM"
```
bill-dall: for all the xRemoteFiles we should be adding this as a parameter:
			MatchSource = $false

Otherwise it will re-download on reboot which we don't want.

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -0,0 +1,210 @@
...
+echo "-->Registering with DNS"
+DOMAIN_UPPER=$(echo "$DOMAIN_NAME" | tr "[a-z]" "[A-Z]")
+IP_ADDRESS=$(hostname -I | grep -Eo '10.([0-9]*\.){2}[0-9]*')
+echo $PASSWORD | sudo kinit $USERNAME@$DOMAIN_UPPER
+touch dns_record
+echo "update add $VM_NAME.$DOMAIN_NAME 600 a $IP_ADDRESS" > dns_record
+echo "send" >> dns_record
+sudo nsupdate -g dns_record
+
+# Install the EPEL repository
```
bill-dall: Does EPEL stand for anything interesting? Why do we need it? (Maybe can flesh out the comment a bit more.)

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -0,0 +1,210 @@
...
+# the third argument is the VM name
+VM_NAME=$3
+# the forth argument is the domain name
+DOMAIN_NAME=$4
+# the fifith argument is the username
+USERNAME=$5
+# the sixth argument is the password
+PASSWORD=$6
+# the seventh argument is the domain group to join
+GROUP=$7
```
bill-dall: It's not joining a domain group though, is that correct?

dev/domain-controller/new-admin-vm/Install-CAM.ps1
jackyycheng: Ok, I will add that to all xRemoteFile

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.sh
jackyycheng: This script was taken from 1 click deployment. I'm not sure why it was installing EPEL repository. @fwang-teradici , Do you know why we need this?

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.sh
jackyycheng: Yes, I thought we could join the domain group but it seems it would require a lot of work to do it.

fwang-teradici: From the document to install PCoIP Agent on REPL
https://techsupport.teradici.com/ics/support/kbanswer.asp?deptID=15164&task=knowledge&questionID=3147

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: From the document to install PCoIP Agent on REPL
https://techsupport.teradici.com/ics/support/kbanswer.asp?deptID=15164&task=knowledge&questionID=3147

fwang-teradici: @bill-dall: EPEL: Extra Packages for Enterprise Linux

bill-dall: Okay thank.

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.sh
bill-dall: Yeah, sorry I should have been thinking about the 'done' criteria more when we were talking a few days ago.  This site seems to have a binary for adtool - from 2009??? https://directory.fsf.org/wiki/Adtool

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -0,0 +1,210 @@
...
+        exit 105
+        ;;       
+esac
+
+# Join domain
+echo "-->Install required packages to join domain"
+sudo yum -y install sssd realmd oddjob oddjob-mkhomedir adcli samba-common samba-common-tools krb5-workstation openldap-clients policycoreutils-python
+sudo systemctl enable sssd
+
+echo "-->Joining the domain"
```
cptera: Could this be a function? And only call it if all Domain Credentials are provided (otherwise don't join domain)?

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -0,0 +1,210 @@
+#!/bin/bash
+
```
bdyck: We should be using `set -eu` in shell scripts to ensure that the script exits on a) uncaught errors and b) if variables are undefined. (For reasons why this is a good idea, google for Valve blowing away user's home directories because of an unset variable in an installer script ;) )

bdyck:
I have a question regarding shell scripting... I see several things that are concerning style-wise and could probably be solved by adopting a style guide (using `local` for local vars in functions, inconsistent or improper quoting, using `[` instead of `[[` for conditionals, writing errors to stdout instead of stderr, etc).

I know that sounds headache-inducing but shell scripting is pretty unforgiving and "style" is probably better defined as "things that will avoid / minimise future bugs". :) As a quick solution, I'd suggest [Google's style guide](https://google.github.io/styleguide/shell.xml) -- it's pretty succinct. Thoughts?

dev/domain-controller/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -0,0 +1,311 @@
...
+# the forth argument is the domain name
+DOMAIN_NAME="$4"
+# the fifith argument is the username
+USERNAME="$5"
+# the sixth argument is the password
+PASSWORD="$6"
+# the seventh argument is the domain group to join
+GROUP="$7"
+# set the domain controller address
+DC_ADDRESS="vm-dc.$DOMAIN_NAME"
```
bill-dall: I think the whole DC address should be a parameter - we want to support corporate domains very soon. Or - I think the Windows machines find the local domain controller through a DNS-srv query or something like that? Because you never have to type in the DC name in Windows I don't think.

jackyycheng:
@bill-dall , before when we did the one click deployment, we skip the desktop initialization part for the user, like choosing desktop language, keyboard setting and etc. Once that whole block is removed, user will have to go through that process.

bill-dall:
Okay thanks Jacky,

That makes sense if the user wants to customize. I'd almost recommend leaving your code in (but commented out) so it's clear how to auto-setup in the future.

bdyck:
@bill-dall That code has bugs in it, which is partly why it was removed -- after talking with @jackyycheng and @fwang-teradici, we decided it would be easier to just remove the functionality than fix it, since it's not critical.

We should be able to just remove it entirely and if we need to resurrect it, Git's history will have it. :)

PR #32, opened 2017-09-29
Tsw 66292 split arm template cleanup

roy-satadru:
Split template deployment cleanup. Note automation test script for cam front end needs to be fixed to point to the right URL for the deployment template. I've changed it in a branch and created a PR so these two PRs would need to be merged at the same time.


frontend/azuredeploy.parameters.json
```diff
@@ -0,0 +1,36 @@
...
+            "value": "GEN-PASSWORD"
+        },
+        "tenantID": {
+            "value": "null"
+        },
+        "CAMDeploymentBlobSource": {
+            "value": "https://teradeploy.blob.core.windows.net/binaries"
+        },
+        "_agentArtifactsLocation": {
+            "value": "https://raw.githubusercontent.com/teradici/deploy/TSW-66292-split-ARM-template-cleanup/end-user-application-machines/new-agent-vm"
```
bill-dall: Same as the other parameters files - I think we should make them all work after being merged to master.

root/azuredeploy.json
```diff
@@ -0,0 +1,319 @@
...
+      }
+    },
+    {
+      "type": "Microsoft.Compute/virtualMachines",
+      "name": "[variables('adVMName')]",
+      "apiVersion": "2015-06-15",
+      "location": "[resourceGroup().location]",
+      "dependsOn": [
+        "[resourceId('Microsoft.Storage/storageAccounts', variables('storageAccountName'))]",
+        "VNet"
```
bill-dall: This needs to depend on the network interface (above) (I just had a failure.)

bill-dall:
New changes look good. You'll have to do a new pull from master to resolve the conflicts before we can approve/merge though I think.

PR #33, opened 2017-10-02
Enabling tlsv1.2

ivpetrov87:


dev/domain-controller/new-admin-vm/Install-CAM.ps1
```diff
@@ -1383,6 +1383,7 @@ brokerLocale=en_US
 "@
 					[System.Net.ServicePointManager]::CertificatePolicy = New-Object TrustAllCertsPolicy
 				}
+                [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```
bill-dall: Do we have the indenting right? I hate indenting...

ivpetrov87: i should have looked at it here... we do not. I will fix it. One is using tabs and the other spaces

PR #36, opened 2017-10-03
* Enabled Linux Auto-Off feature

cptera:


end-user-application-machines/new-agent-vm/rhel-standard-agent.json
```diff
@@ -251,6 +252,27 @@
                 "commandToExecute": "[concat('bash ', variables('scriptFileName'), ' \"', parameters('registrationCode'), '\" \"', parameters('agentType'), '\" \"', variables('vmSettings').vmNamePrefix, '\" \"', parameters('domainToJoin'), '\" \"', parameters('domainUsername'), '\" \"', parameters('domainPassword'), '\" \"', parameters('domainGroupToJoin'), '\"')]"
                 }
               }
-            }
+            },
+            {
```
bill-dall: I didn't know you could run two script extensions - I guess that's new? Cool.

end-user-application-machines/new-agent-vm/Install-Idle-Shutdown.sh
```diff
@@ -0,0 +1,280 @@
...
+# there are instances of pcoip-server processes running.
+
+# Global settings
+#
+
+# minimum number of minutes machine must be idle before shutting down
+MIN_IDLE_WAIT_TIME=5
+
+# maximum number of minutes machine must be idle before shutting down
+MAX_IDLE_WAIT_TIME=10000 # about 7 days
```
cptera: the idle time period and cpu limit and polling interval has a min-max range and we use these to validate and set user provided values to them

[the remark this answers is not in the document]
cptera: the idle time period and cpu limit and polling interval has a min-max range and we use these to validate and set user provided values to them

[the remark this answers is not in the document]
bill-dall: Ahh- so if the feature is enabled the maximum wait time that can be specified is about 7 days. Makes sense. You might want to chagne the comment a bit.

end-user-application-machines/new-agent-vm/Install-Idle-Shutdown.sh
```diff
@@ -0,0 +1,282 @@
...
+    fi
+    echo 0
+}
+
+####################################################################
+# Returns an integer representing the percentage cpuload of the busiest
+# cpu.  e.g. 10 = 10%
+# We do not attempt to normalize
+####################################################################
+function getCPULoad() {
```
bdyck: I'm curious why this function exists -- reading `/proc/loadavg` should produce what we need?

end-user-application-machines/new-agent-vm/Install-Idle-Shutdown.sh
```diff
@@ -0,0 +1,282 @@
...
+}
+
+main $@
+EOF
+}
+
+function install() {
+    echo "Installing"
+    touch ${SERVICE_PATH}
+    chmod 664 ${SERVICE_PATH}
```
bdyck: I'm curious why we're chmod'ing to 664 instead of 644 here; files -- especially system config files -- typically aren't group-writable.

end-user-application-machines/new-agent-vm/Install-Idle-Shutdown.sh
```diff
@@ -0,0 +1,282 @@
...
+    chmod 664 ${SERVICE_CONFIG}
+    create_monitor_script
+    cat <<EOF> ${SERVICE_PATH}
+[Unit]
+Description=Teradici CAM Idle Shutdown monitoring service
+After=pcoip.service
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/sudo /usr/bin/bash ${MONITOR_SCRIPT} \${MinutesIdleBeforeShutdown} \${PollingIntervalMinutes} \${CPUUtilizationLimit}
```
bdyck: The service config doesn't specify a user to run this service as, which means the service will run under the logged-in user; since it requires sudo, that means a) the logged-in user requires not just sudo permission but passwordless sudo.

I think it would be better to add a `User` directive to the service config and have this run as root.

end-user-application-machines/new-agent-vm/Install-Idle-Shutdown.sh
```diff
@@ -0,0 +1,282 @@
...
+    echo 0
+    return
+}
+
+####################################################################
+# convert string to integer, 0 if it is invalid
+####################################################################
+function validateInt(){
+    typeset -r input=$1
+    if [[ $input =~ ^[0-9]+$ ]]; then
```
bdyck: Hmm. Both styles of test syntax seem to be used in this script (`[` and `[[`). I think `[[` is preferred as it [reduces errors](https://google.github.io/styleguide/shell.xml?showone=Test,_%5B_and_%5B%5B#Test,_%5B_and_%5B%5B).

end-user-application-machines/new-agent-vm/Install-Idle-Shutdown.sh
```diff
@@ -0,0 +1,282 @@
...
+    if [ $result -gt 100 ]; then
+        result=100
+    fi
+    echo $result
+}
+
+####################################################################
+# Set the globals wait_minutes and poll_interval
+####################################################################
+function processArgs () {
```
bdyck: I'm not sure I understand why the script accepts these values as command-line args when we are only running it in a non-interactive context. If the interactivity isn't required, we could remove the entire polling loop and simplify the script by using a [systemd.timer unit](http://0pointer.de/public/systemd-man/systemd.timer.html) to periodically execute the corresponding service.

end-user-application-machines/new-agent-vm/Install-Idle-Shutdown.sh
cptera: ok i can change that

end-user-application-machines/new-agent-vm/rhel-standard-agent.json
cptera: Turns out you can't, I had to switch this to CustomScriptForLinux so it wouldn't clash with the other one

end-user-application-machines/new-agent-vm/Install-Idle-Shutdown.sh
```diff
@@ -1,257 +1,271 @@
...
+# File containing state information
+startTimeFile = "/tmp/CAMIdleState"
+cpuUsageFile = "/tmp/CAMIdleStateCPU"
+
+def getStartTime():
+    """
+    Check what the start time is
+    """
+    if os.path.exists(startTimeFile):
+        f = open(startTimeFile, 'rb')
```
bdyck: It's a good idea to use a context block for file I/O, like this:
```python
with open(startTimeFile, 'rb') as f:
    startTime = f.read()
```
The context block takes care of the underlying file handle management and keeps things neat (i.e. flushing when you write, ensuring that it's closed when the block exits regardless of whether errors occur). :)

end-user-application-machines/new-agent-vm/Install-Idle-Shutdown.sh
```diff
@@ -1,257 +1,271 @@
...
+            elapsedTime = currentTime - startTime
+            syslog.syslog("CPU usage is now {cpu}%, it is below idle threshold. CPU has been idle for at least {min} minutes".format(
+                min=int(elapsedTime/60),
+                cpu=cpuUsage
+            ))
+            if elapsedTime > waitTime:
+                syslog.syslog("CPU has been idle for more than {} minutes, shutting down".format(
+                    waitTime
+                ))
+                os.system('/sbin/shutdown')
```
bdyck: `os.system()` is deprecated; this should be able to be replaced by `subprocess.Popen(['/sbin/shutdown'])`. In addition, `clearStartTime()` should be called beforehand -- otherwise, that line will never be executed because script execution will be terminated by the call to `shutdown`.

PR #39, opened 2017-10-03
Change application VM size from D2_v3 to D2_v2

roy-satadru:
Revert VM size back to D2_v2

### DIFF
--- a/azuredeploy.json
+++ b/azuredeploy.json
@@ -215,7 +215,7 @@
               "value": "[reference('Root').outputs.adminDesktopVMName.value]"
             },
           "vmSize": {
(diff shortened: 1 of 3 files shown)


azuredeploy.json
```diff
@@ -215,7 +215,7 @@
               "value": "[reference('Root').outputs.adminDesktopVMName.value]"
             },
           "vmSize": {
-            "value": "Standard_D2_v3"
+            "value": "Standard_D2_v2"
```
bill-dall: This is necessary but I don't think sufficient. If you do a whole tree search I think you find _v3 in three locations?

PR #45, opened 2017-10-19
Private blob storage

ivpetrov87:


end-user-application-machines/new-agent-vm/Install-PCoIPAgent.ps1
```diff
@@ -18,6 +18,9 @@ Configuration InstallPCoIPAgent
         [String]$sumoCollectorID,
 
         [Parameter(Mandatory=$false)]
+        [String]$sasToken,
```
bill-dall: Should we make this a credential populated by a SecureString in the JSON like we're handling other sensitive data?

end-user-application-machines/new-agent-vm/rhel-standard-agent.json
```diff
@@ -246,7 +246,7 @@
                 "typeHandlerVersion": "2.0",
                 "autoUpgradeMinorVersion": true,
                 "settings": {
-                    "fileUris": [ "[concat(parameters('_artifactsLocation'), '/', variables('scriptFileName'))]" ]
+                    "fileUris": [ "[concat(parameters('_artifactsLocation'), '/', variables('scriptFileName'), parameters('_artifactsLocationSasToken'))]" ]
```
bill-dall: That's interesting - I wonder if this works under 'protectedsettings' as well now that we have the token.

end-user-application-machines/new-agent-vm/rhel-standard-agent.json
```diff
@@ -267,12 +267,12 @@
                 "typeHandlerVersion": "1.4",
                 "autoUpgradeMinorVersion": true,
                 "settings": {
-                    "fileUris": [ "[concat(parameters('_artifactsLocation'), '/', variables('idleShutdownScriptFileName'))]" ]
+                    "fileUris": [ "[concat(parameters('_artifactsLocation'), '/', variables('idleShutdownScriptFileName'), parameters('_artifactsLocationSasToken'))]" ]
```
bill-dall: Same here.

frontend/new-admin-vm/Install-CAM.ps1
```diff
@@ -210,6 +222,48 @@ Configuration InstallCAM
 			MatchSource = $false
 		}
 
+		xRemoteFile Download_Install_Agent.ps1
```
bill-dall: When do we use the 'naked' ps1 file? Maybe I'll see later.

frontend/new-admin-vm/Install-CAM.ps1
```diff
@@ -976,6 +1037,67 @@ graphURL=https\://graph.windows.net/
...
+				try {
+						Get-AzureStorageContainer -Name $container_name -Context $ctx -ErrorAction Stop
+				} Catch {
+						New-AzureStorageContainer -Name $container_name -Context $ctx -Permission "Blob"
+				}
+				Write-Host "Uploading files to private blob"
+				ForEach($file in $new_agent_vm_files) {
+						$filepath = Join-Path $using:LocalDLPath $file
+						try {
+							Get-AzureStorageBlob -Context $ctx -Container $container_name -Blob "remote-workstation\$file" -ErrorAction Stop
```
bill-dall: Backslash or forward slash?

frontend/new-admin-vm/Install-CAM.ps1
```diff
@@ -976,6 +1037,67 @@ graphURL=https\://graph.windows.net/
...
+				Set-AzureKeyVaultSecret -VaultName $kvName -Name $storageAccountSecretName -SecretValue (ConvertTo-SecureString $acct_name -AsPlainText -Force) -ErrorAction stop
+				$storageAccountKeyName = "userStorageAccountKey"
+				Set-AzureKeyVaultSecret -VaultName $kvName -Name $storageAccountKeyName -SecretValue (ConvertTo-SecureString $acctKey -AsPlainText -Force) -ErrorAction stop
+
+				$saSasToken = New-AzureStorageContainerSASToken -Name $container_name -Permission rwdl -Context $ctx
+				$saSasTokenSecretName = 'userStorageAccountSaasToken'
+				Set-AzureKeyVaultSecret -VaultName $kvName -Name $saSasTokenSecretName -SecretValue (ConvertTo-SecureString $saSasToken -AsPlainText -Force) -ErrorAction stop
+
+				# using the blob uri + the token from the key vault will allow the web interface to retrieve required information from private blob
+				[System.Environment]::SetEnvironmentVariable("CAM_KEY_VAULT_NAME", $kvName, "Machine")
```
bill-dall: maybe add a note that the environment variable setting doesn't take effect in the current process.

end-user-application-machines/new-agent-vm/rhel-standard-agent.json
ivpetrov87: i got this from examples online but good point, i'll try it from protected settings

frontend/new-admin-vm/Install-CAM.ps1
ivpetrov87: at the moment we don't, my understanding was the we may potentially allow customers to customize this and rebuild zip file since its in their blob. If this is not correct, then i will remove it.

frontend/new-admin-vm/Install-CAM.ps1
ivpetrov87: Backlash works. They make it look like the windows file system. Although forward slash should also work

frontend/new-admin-vm/Install-CAM.ps1
ivpetrov87: Good point

end-user-application-machines/new-agent-vm/Install-PCoIPAgent.ps1
ivpetrov87: I'll look into how to do that.  i don't think we are passing this anywhere else to the DSC. 

We pass the registration code as PSCredential but in this case i don't know where to cast it as pscredential

frontend/new-admin-vm/Install-CAM.ps1
bill-dall: No real harm in copying I guess. Just making sure I knew where it was used :)

frontend/new-admin-vm/Install-CAM.ps1
bill-dall: For some anal-retentive reason I'd prefer forward slash here as I expect that's how most 'blob fake filesystems' are done. Or did you see it that way in a bunch of examples?

ivpetrov87:
@bill-dall 
Testing it now with a new deployment. will double check that all the parameters get added to environment variables. (And i forgot to add the note about those, will add it now)

end-user-application-machines/new-agent-vm/rhel-standard-agent.json
```diff
@@ -99,7 +99,7 @@
             "metadata": {
                 "description": "Auto-generated token to access _artifactsLocation"
             },
-            "defaultValue": ""
+            "defaultValue": "?public"
```
cptera: is the '?' a typo?

end-user-application-machines/new-agent-vm/Install-PCoIPAgent.ps1
```diff
@@ -88,9 +88,12 @@ Configuration InstallPCoIPAgent
                 $installerFileName = "SumoCollector_windows-x64_19_182-25.exe"
                 $uninstallerRegistryID = "7857-4527-9352-4688"  # This will need to change for every installer version change!
 
+				$sasToken = ($using:sasTokenAsCred).GetNetworkCredential().password
+				$blobLocation = ($using:sasTokenAsCred).GetNetworkCredential().username
+
                 $sumo_package = "https://teradeploy.blob.core.windows.net/binaries/$installerFileName"
-                $sumo_config = "$using:gitLocation/sumo.conf"
-                $sumo_collector_json = "$using:gitLocation/sumo-agent-vm.json"
+                $sumo_config = "$blobLocation/sumo.conf$sasToken"
```
cptera: not a big deal if it works as is, but i think "${blobLocation}/sumo.conf${sasToken}" would be better

frontend/new-admin-vm/Install-CAM.ps1
```diff
@@ -210,6 +222,48 @@ Configuration InstallCAM
...
+		xRemoteFile sumo 
+		{
+				Uri = "$gitLocation/$sumoConf"
+				DestinationPath = "$LocalDLPath\$sumoConf"
+				MatchSource = $false
+		}
+
+		xRemoteFile Download_Install_Agent.ps1.zip
+		{
+				Uri = "https://teradeploy.blob.core.windows.net/binaries/Install-PCoIPAgent.ps1.zip"
```
cptera: The blob url is defined as a parameter (sourceUri I think)

frontend/new-admin-vm/Install-CAM.ps1
```diff
@@ -976,6 +1037,69 @@ graphURL=https\://graph.windows.net/
...
+						}
+				}
+
+				$blobUri = (((Get-AzureStorageBlob -Context $ctx -Container $container_name)[0].ICloudBlob.uri.AbsoluteUri) -split '/')[0..4] -join '/'
+
+				# this is the url to access the blob account
+				$blobUriSecretName = "userStorageAccountUri"
+				Set-AzureKeyVaultSecret -VaultName $kvName -Name $blobUriSecretName -SecretValue (ConvertTo-SecureString $blobUri -AsPlainText -Force) -ErrorAction stop
+
+				$storageAccountSecretName = "userStorageName"
```
cptera: Do we want to use the word "user" for this? "user" sounds like end-user, but this isn't really end-user storage. maybe CAMStorageName?

ivpetrov87: if you think about it though all the templates stored there are for the end-user. The user customizes them and stores them there, so i named it userStorage :)

end-user-application-machines/new-agent-vm/rhel-standard-agent.json
ivpetrov87: not a typo, I need to provide a default value or the credential wont get serialized.

If i jsut leave it as public, it gets appended to the url and it fails the very first deployment because it uses github.com and appends "public" to the url which cases 404.

If we append ?public or really any query string, that does not cause a 404

frontend/new-admin-vm/Install-CAM.ps1
ivpetrov87: changed

end-user-application-machines/new-agent-vm/Install-PCoIPAgent.ps1
ivpetrov87: changed it, you are right its clearer

azuredeploy.json
```diff
@@ -247,6 +247,9 @@
           "CAMDeploymentBlobSource": {
             "value": "[parameters('CAMDeploymentBlobSource')]"
           },
+          "binaryLocation": {
```
cptera: is binary location a different thing from CAMDeploymentBlobSource?

bill-dall:
Merging branch where all binaries are copied to customer blob storage first, to prevent our changes from breaking them later.

PR #46, opened 2017-10-24
TSW-67105 document customer AD setup in domain.md

fwang-teradici:


domain.md
```diff
@@ -0,0 +1,71 @@
+In order to deploy CAM that is integrated with customer own Active Directory (AD), the folloing requirements must be met. When deploying CAM, the **account credential and DC fqdn** must be provided, and the DC must be able to communicated with the given **fqdn**.
+
+# 1) Creating an account with delegated permissions to join a computer to domain. 
+
+The following steps setup an account with delegated privileges to join a computer to domain, assuming an account already created, otherwise, please create an account first.
+
+ * Login Domain Controller Server.
+ * Lanuch Active Directory Users and Computers.
+ * Locate the **OU** that you want to join a computer to and right-click and then click Delegate Control.
```
bill-dall: OU's are great but what should they do before we support OU's? Maybe the doc here is fine if we know what to do in the meantime - can you click on something more 'root' than the OU?

domain.md
```diff
@@ -0,0 +1,71 @@
...
+The following steps setup an account with delegated privileges to join a computer to domain, assuming an account already created, otherwise, please create an account first.
+
+ * Login Domain Controller Server.
+ * Lanuch Active Directory Users and Computers.
+ * Locate the **OU** that you want to join a computer to and right-click and then click Delegate Control. 
+ * The **Delegation of Control Wizard** opens, click **Next**
+ * In the **Users or Groups** page, click **Add** to add a specific user or a specific group to the **Selected users and groups** list, and then click **Next**
+ * In the **Tasks to Delegate** page, click **Create a custom task to delegate**, and then click **Next**.
+ * In the **Active Directory Object Type** page, Click **Only the following objects in the folder**, and then from the list, click to select the **Computer objects** check box. Then, select the check boxes below the list, **Create selected objects in this folder** and **Delete selected objects in this folder**, Click **Next**.
+ * In the **Permissions** page, click to select the following permission check boxes from list:
```
bill-dall: Which one allows unlimited domain joins?

domain.md
```diff
@@ -0,0 +1,71 @@
...
+
+The following steps enable remote powershell admin
+
+* Login Domain Controller Server.
+* Start Windows PowerShell as an administrator.
+* Check the configuration of WinRM service by running following command:  
+  **Get-WmiObject Win32_Service | Where-Object { $_.Name -eq 'WinRM'} | Format-Table**  
+  The value of the StartMode property in the output should be "Auto" and the value of the Status property in the output should be "Running"
+* Set the StartMode property of WinRM service to Automatic by running following command if needed:  
+  **Set-Service WinRM -StartupType Automatic**
```
bill-dall: Online you see a bunch of people recommending winrm-quickconfig. But it seems to enable http by default (yech) - did you look at that?

domain.md
```diff
@@ -0,0 +1,71 @@
...
+  **Restart-Service -Name WinRM**
+* run command to enable remote powershell: **Enable-PSRemoting -Force**
+
+
+# 3) Enable LDAPs on a on Domain Controller (DC) machine. 
+    
+Basically, there are two methods of enabling LDAPS on a DC.
+
+* Method #1: install an Enterprise Root CA  
+  The first method is the easiest: LDAPS is automatically enabled when you install an Enterprise Root CA on a Domain Controller. If you install the AD-CS role and specify the type of setup as ��Enterprise�� on a DC, all DCs in the forest will be automatically be configured to accept LDAPS.
```
bill-dall: Yeah this is what we do for the current DC

domain.md
```diff
@@ -0,0 +1,71 @@
...
+# 3) Enable LDAPs on a on Domain Controller (DC) machine. 
+    
+Basically, there are two methods of enabling LDAPS on a DC.
+
+* Method #1: install an Enterprise Root CA  
+  The first method is the easiest: LDAPS is automatically enabled when you install an Enterprise Root CA on a Domain Controller. If you install the AD-CS role and specify the type of setup as ��Enterprise�� on a DC, all DCs in the forest will be automatically be configured to accept LDAPS.  
+    
+  
+* Method #2: add a Digital Certificate on DC  
+  Requirements for an LDAPS certificate:
```
bill-dall: Is this from somewhere or did you assemble this from a bunch of snippets? If  there's a good MSFT/other doc that has most of the same info, would be good to reference it as well.

domain.md
```diff
@@ -0,0 +1,71 @@
...
+      Start Microsoft Management Console (MMC).  
+      Add the Certificates snap-in that manages certificates on the local computer.  
+      Expand Certificates (Local Computer), expand Personal, and then expand Certificates.  
+      A new certificate should exist in the Personal store. In the Certificate Properties dialog box, the intended purpose displayed       is Server Authentication. This certificate is issued to the computer's fully qualified host name.
+  6)  Enable the host machine account to access to the private key.  
+      The private key will be present in the following location C:\ProgramData\Microsoft\Crypto\Keys\<UniqueContainerName>  
+      Right Click the private key path and click properties --> Security and add read permissions for **NETWORK SERVICE**.
+  
+  Verifying an LDAPS connection:
+  1)  Start the Active Directory Administration Tool (Ldp.exe).
```
bill-dall: Where is this available by default? How do you get it?

domain.md
fwang-teradici: yes, in the GUI, you could click on 'root' instead of one of OU. Based my understanding, it does not impact CAM deploy. If the delegation is under some 'OU', then when joining computer to domain using this account, the computer should join under this 'OU' automatically. CAM does not to know which OU the computer will join.

domain.md
fwang-teradici: Once delegated permission assign to the account, the limitation is removed

domain.md
fwang-teradici: I searched a lot online, basically it looks the same. Create a CA with specially requirement. and install to the machine, giving the permission to access private key.

domain.md
fwang-teradici: in AD DC machine, once DS role installed, it is default.

domain.md
fwang-teradici: In azure VM, I use Enable-PSRemoting –force command. From my machine, I have to use https to connect, otherwise it failed.  If it is in the trusted domain, it might use http, it not, you must either use HTTPS as the transport or add the remote machine to the TrustedHosts configuration settings

domain.md
bill-dall: Oh that's cool. Maybe you just solved another problem, which is great! :) But if 'root' is okay (what's it called in the GUI?) then would be good to say that.

domain.md
bill-dall: Great!

domain.md
bill-dall: Okay. Yeah good to confirm that http doesn't work.

bill-dall: Or maybe that's incorrect - if both client and server are part of the same (or trusted) domain then they might negotiate different transport security than https, using Kerberos or CredSSP or something. So it's more just knowing that we haven't enabled 'cleartext' transport. Not sure how to verify that. Like, why would winrm-quickconfig enable http? Seems ludicrous in today's day and age unless there's some other trick we're missing.

domain.md
bill-dall: So you call ldp.exe to localhost:636 then?

fwang-teradici: I have not try yet, but ldp.exe is there

domain.md
fwang-teradici: Active Directory Users and Computers

domain.md
cptera: ¡°Enterprise¡± ?

domain.md
```diff
@@ -0,0 +1,71 @@
...
+    
+Basically, there are two methods of enabling LDAPS on a DC.
+
+* Method #1: install an Enterprise Root CA  
+  The first method is the easiest: LDAPS is automatically enabled when you install an Enterprise Root CA on a Domain Controller. If you install the AD-CS role and specify the type of setup as ��Enterprise�� on a DC, all DCs in the forest will be automatically be configured to accept LDAPS.  
+    
+  
+* Method #2: add a Digital Certificate on DC  
+  Requirements for an LDAPS certificate:  
+  1) Digital Certificate must be valid for the purpose of ��Server Authentication.�� This means that they must contain the Server Authentication object identifier (OID). OIDs are like the Internet domain name space. They are series of numbers separated by dots, each with a specific meaning. For this purpose, the relevant OID we��re looking for is 1.3.6.1.5.5.7.3.1.
```
cptera: ¡°Server Authentication.¡± ?

cptera: we¡¯re

domain.md
```diff
@@ -0,0 +1,71 @@
...
+    
+  
+* Method #2: add a Digital Certificate on DC  
+  Requirements for an LDAPS certificate:  
+  1) Digital Certificate must be valid for the purpose of ��Server Authentication.�� This means that they must contain the Server Authentication object identifier (OID). OIDs are like the Internet domain name space. They are series of numbers separated by dots, each with a specific meaning. For this purpose, the relevant OID we��re looking for is 1.3.6.1.5.5.7.3.1.  
+  2) The Subject name or the first name in the Subject Alternative Name (SAN) must match the Fully Qualified Domain Name (FQDN) of the host machine, such as Subject:CN=server.domain.com.  
+  3) The host machine account needs to have access to the private key. This is done when the digital certificate request is issued from that machine, or when the private key was exported and imported to a different machine.  
+    
+  The following steps setup LDAPS:  
+  1)  Create the request file.
```
cptera: How do you do this?

fwang-teradici: there are different tool to create certificate request file.  Windows has certreq tool, Java has keytool

domain.md
fwang-teradici: unicode issue. correct them

domain.md
bill-dall: "Create the certificate request (usually PEM) file with your usual certificate generation tool." ? Something like that?

fwang-teradici: yes, 
Java: https://docs.oracle.com/cd/E19509-01/820-3503/ggezu/index.html
openssl: https://www.ssl.com/how-to/manually-generate-a-certificate-signing-request-csr-using-openssl/
https://blogs.technet.microsoft.com/pki/2009/08/05/how-to-create-a-web-server-ssl-certificate-manually/

PR #54, opened 2017-11-20
Accept ext cert

fwang-teradici:


Deploy-CAM.ps1
```diff
@@ -1201,7 +1280,7 @@ function Deploy-CAM()
 	Write-Host "Using temporary directory $tempDir for intermediate files"
 	if(-not (Test-Path $tempDir))
 	{
-		New-Item $tempDir -type directory > $null
+		New-Item $tempDir -type directory | Out-Null
```
bill-dall: It definitely doesn't matter in this case but I found this posting interesting re. performance of the various 'don't pollute the pipeline' strategies :https://stackoverflow.com/questions/5260125/whats-the-better-cleaner-way-to-ignore-output-in-powershell

Deploy-CAM.ps1
```diff
@@ -129,8 +137,8 @@ function Register-CAM()
 	)
 
     $camDeploymentInfo = @{} #start with an empty hash map
-	$camRegistrationError = ""
```
bill-dall: But it's used outside of the for loop below... Why the change?

Deploy-CAM.ps1
```diff
@@ -331,8 +339,9 @@ function Register-RemoteWorkstation()
 		subscriptionId = $subscription
 	}
 
-	$machineRegistrationError = ""
 	for($idx = 0; $idx -lt $retryCount; $idx++) {
+		$machineRegistrationError = ""
```
bill-dall: Are we sure that it's not used outside of the for loop as well? (I didn't check.)

Deploy-CAM.ps1
```diff
@@ -905,54 +982,48 @@ function New-PopulatedKeyvault()
 		#generate pfx file from certificate
 		$certPath = $certLoc + '\' + $cert.Thumbprint
 
-		$certPfx = Join-Path $tempDir "self-signed-cert.pfx"
+		if (-not $tempDir) {
```
bill-dall: This is okay to check - but just wondering - Did you hit this case?

Deploy-CAM.ps1
```diff
@@ -905,54 +982,48 @@ function New-PopulatedKeyvault()
...
+			$tempDir = $env:TEMP
+		}
 
-		#generate password for pfx file
-		$certPswd = -join ((48..57) + (65..90) + (97..122) | Get-Random -Count 16 | % {[char]$_})
-		$secureCertPswd = ConvertTo-SecureString -String $certPswd -AsPlainText -Force
+		$certificateFile = Join-Path $tempDir "self-signed-cert.pfx"
+        if (Test-Path $certificateFile) {
```
bill-dall: Yeah that's one reason I wanted to have things happen in a random sub-directory.

Deploy-CAM.ps1
```diff
@@ -905,54 +982,48 @@ function New-PopulatedKeyvault()
...
-	finally {
-		#done with files
-		if(Test-Path $certPfx) { Remove-Item $certPfx -ErrorAction SilentlyContinue }
-		if(Test-Path $certPath) { Remove-Item $certPath -ErrorAction SilentlyContinue }
-	}
-	return $secretHash
-}
 
+	#delete certificate file if it is generated
+	if ($needToCreateSelfCert -and	(Test-Path  $certificateFile) ) {
```
bill-dall: This used to be in a 'finally' block to ensure it's always (or rather - more often) - properly cleaned up. Does that make sense in this case?

Deploy-CAM.ps1
fwang-teradici: if it is outside of the loop, for example, at first iteration, there is error and it get the value, at the second iteration, it processed successful. Exit the loop, because the value is not clear. so the if statement if($camRegistrationError) {throw $camRegistrationError} executed and it throw exception. This will break your initial purpose of the retrying loop

Deploy-CAM.ps1
fwang-teradici: when I run this function in unit, I hit this case, so just add this in case

Deploy-CAM.ps1
fwang-teradici: To remove the duplicated the code to read certificate file in both cases which are generating on the fly or providing certificate file. the deleting code is outside of try block and it only delete the generating certificate file.

Deploy-CAM.ps1
fwang-teradici: Using pipeline Out-Null purpose is to wait for the cmdlet complete finish. If this is not the case, then do not using pipepline Out-Null here. That is why I use Out-Null instead of $null.

Deploy-CAM.ps1
bill-dall: Got it.

Deploy-CAM.ps1
bill-dall: Okay. Cool. I wasn't thinking about waiting.

Deploy-CAM.ps1
fwang-teradici: the same as above

Deploy-CAM.ps1
bill-dall: Good catch! I think I'd leave both in though. I guess it doesn't really matter and maybe 'my C++' is showing but would be nice to have the variable declared/initialized outside of the loop if it's consumed outside of the loop. A style thing that wouldn't affect functionality in this case... Thoughts?

fwang-teradici: I would declare the variable outside the loop and reset the value in the loop. so make the code more clear. No sure the powershell variable scope.

bill-dall: Sounds good. variable scope is 'within function' for local variables so variables within loops persist past the loops, unlike C++ or many other languages.

PR #55, opened 2017-11-21
Removed some dead code, separated CAM deployment structure population…

bill-dall:
… to a separate function, single-sourced more strings.

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -281,136 +281,6 @@ public class TrustAllCertsPolicy : ICertificatePolicy {
 
 
 
-function Register-RemoteWorkstation()
-{
-	Param(
-		[bool]
-		$verifyCAMSaaSCertificate = $true,
-		
-		# Retry for Remote Workstation Registration
-		$retryCount = 3,
-		$retryDelay = 10,
-
-		[parameter(Mandatory=$true)] 
-		$subscription,
-		
-		[parameter(Mandatory=$true)]
-		$client,
-		
-		[parameter(Mandatory=$true)]
-		$key,
-		
-		[parameter(Mandatory=$true)]
-		$tenant,
-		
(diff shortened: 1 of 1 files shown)


bill-dall:
When I was flipping through the file yesterday I decided to so some cleanups for maintainability.

Deploy-CAM.ps1
```diff
@@ -1095,6 +967,77 @@ function New-CAMAppSP()
...
 }
 
+#creates cam deployment info structures and populates keyvault
+function New-CamDeploymentInfo()
+{
+	param(
+	    [parameter(Mandatory=$true)] 
+	    $subscriptionId, # Azure subscription being operated on
+
+		[parameter(Mandatory=$true)]
```
ivpetrov87: is this a mixed space - tab indent issue?

Deploy-CAM.ps1
```diff
@@ -1186,6 +1129,11 @@ function Deploy-CAM()
 	$CAMConfig.internal.rwLocalSecretName = 'remoteWorkstationLocalAdminPassword'
 	$CAMConfig.internal.csLocalSecretName = 'connectionServiceLocalAdminPassword'
 	$CAMConfig.internal.saSasTokenSecretName = "userStorageAccountSaasToken"
+	$CAMConfig.internal.blobUriSecretName = "userStorageAccountUri"
```
ivpetrov87: If this is how we are saving them into key vault, i think we should let primesoft know that the instructions for retrieving this information for the front end should probably change, the examples i gave were using these as separate secrets

Deploy-CAM.ps1
bill-dall: Yeah - I've got some more edits coming in a subsequent PR. I'll look at unifying on I guess tabs in this case based on the editor defaults, for that PR.

PR #59, opened 2017-12-01
update description for AgentChannel

fwang-teradici:


fwang-teradici:
update the description for AgentChannel parameter

PR #60, opened 2017-12-01
Deploy a connection service in its own resource group.

bill-dall:
this seems to work...

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -1141,7 +1141,6 @@ function Generate-CamDeploymentInfoParameters {
 # Deploy a connection service over a current deployment
 function New-ConnectionServiceDeployment() {
     param(
-        $CSsuffix,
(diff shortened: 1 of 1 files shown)


Deploy-CAM.ps1
```diff
@@ -1152,121 +1151,46 @@ function New-ConnectionServiceDeployment() {
...
-					"id": "$kvID"
-				},
-				"secretName": "artifactsLocation"
-			}
-		}
-	}
-}
-"@
+
+    $csRGName = $null
```
cptera: Maybe add a comment for what the while loop does?

cptera:
Wow, that was a pretty straight forward change. Looks good to me

Deploy-CAM.ps1
bill-dall: Good point. Done (in my other working branch.)

PR #61, opened 2017-12-04
Bddc2b

bill-dall:
This is the 3 resource group change to make it even easier to deploy new connection services. Let's use this (or a reviewed version of this :) as a basis of future work. The only issue I see that I'm concerned about right now is an increased occurrence of 'can't find LDAP certificate' which needs investigation but I believe is independent of the changes in here.


Deploy-CAM.ps1
```diff
@@ -35,7 +35,7 @@ param(
     $certificateFilePassword = $null,
 
     $camSaasUri = "https://cam-antar.teradici.com",
-    $CAMDeploymentTemplateURI = "https://raw.githubusercontent.com/teradici/deploy/bddc2/azuredeploy.json",
+    $CAMDeploymentTemplateURI = "https://raw.githubusercontent.com/teradici/deploy/bddc2b/azuredeploy.json",
```
bill-dall: revert...

Deploy-CAM.ps1
```diff
@@ -1152,6 +1132,10 @@ function New-ConnectionServiceDeployment() {
     # TODO - make sure user account has keyvault secret access here. Try to add self if not
     # and if doesn't work, probably fail.
 
+    # Find a connection service resouce group name that can be used.
```
bill-dall: resouce...

connection-service/new-admin-vm/new-admin-vm.json
```diff
@@ -72,11 +72,11 @@
                 "typeHandlerVersion": "2.9",
                 "autoUpgradeMinorVersion": true,
                 "settings": {
-                    "modulesUrl": "[concat(parameters('CAMDeploymentBlobSource'),'/Install-CAM-CS.ps1.zip')]",
+                    "modulesUrl": "[concat(parameters('CAMDeploymentBlobSource'),'/Install-ConnectionServer.ps1.zip')]",
```
bill-dall: I changed the name (again) mostly to avoid having to create a new blob storage. But we do have to stick with a name in the end - what do you think of this one?

fwang-teradici: I prefer to Install-ConnectionServer.ps1.zip

bill-dall:
To start collapsing the large number of branches going on here I'm going to merge this one.

PR #62, opened 2017-12-06
Cloudshell deploy over existing dc vnet

bill-dall:


Deploy-CAM.ps1
```diff
@@ -39,6 +39,26 @@ param(
...
+    [String]
+    $vnetName,
+
+    [parameter(Mandatory=$false)]
+    [String]
+    $GatewaySubnetName,
+
+    [parameter(Mandatory=$false)]
+    [String]
+    $CloudServicesSubnetName,
```
bill-dall: CloudServicesSubnetName <- singular or plural? (for the service(s)

Deploy-CAM.ps1
```diff
@@ -1008,7 +1028,7 @@ function New-CAMAppSP() {
 
     # Get tenant ID for this subscription
     $subForTenantID = Get-AzureRmContext
-    $tenantID = $subForTenantID.Tenant.Id
+    $tenantID = $subForTenantID.Tenant.TenantId
```
bill-dall: Were you having an error here? I can't remember which version is 'the future'...

Deploy-CAM.ps1
```diff
@@ -1172,8 +1192,15 @@ function New-ConnectionServiceDeployment() {
         $rg = Get-AzureRmResourceGroup -ResourceGroupName $csRGName -ErrorAction SilentlyContinue
         if($rg)
         {
-            # found the resource group - do the loop with an incremented number try to find a free name
-            $csRGName = $null
+            # Check if Resource Group is empty
+            $Resources = Find-AzureRmResource -ResourceGroupNameEquals $csRGName
+            $outputstr = $RG.ResourceGroupName + " - " + $Resources.Length
+            Write-Output $outputstr
```
bill-dall: This a debug print? (Also note that Write-Output populates the object pipeline where Write-Host doesn't, I think.)

Deploy-CAM.ps1
```diff
@@ -1172,8 +1192,15 @@ function New-ConnectionServiceDeployment() {
...
+            if( -not $Resources.Length -eq 0)
+            {
+                # found the resource group - do the loop with an incremented number try to find a free name
```
bill-dall: found that the resource group was not empty?

Deploy-CAM.ps1
```diff
@@ -1789,29 +1829,53 @@ function Deploy-CAM() {
...
+                -keyVault $CAMRootKeyvault `
+                -testDeployment $testDeployment `
+                -tempDir $tempDir
+        }
+        else
+        {
+            
 
-        $generatedDeploymentParameters = @"
+            # keyvault ID of the form: /subscriptions/$subscriptionID/resourceGroups/$azureRGName/providers/Microsoft.KeyVault/vaults/$kvName
```
bill-dall: Should we make the whole 'else' thing here a new function?

Deploy-CAM.ps1
```diff
@@ -2202,7 +2268,86 @@ else {
...
+            if ( -not ($vnet.Subnets | ?{$_.Name -eq $vnetConfig.RWsubnetName}) ) {
+                # Does not exist
+                Write-Host ("{0} not found in root resource group VNet {1}" -f ($vnetConfig.RWsubnetName,$vnet.Name))
+                $vnetConfig.RWsubnetName = $null
+            }
+        } while (-not $vnetConfig.RWsubnetName)
+    } else {
+        #create new DC and vnets
+        if( -not $vnetConfig.vnetName ) {
+            $vnetConfig.vnetName = "vnet-CloudAccessManager"
```
bill-dall: I think I'd go with the original names but happy to discuss.

Deploy-CAM.ps1
```diff
@@ -2202,7 +2268,86 @@ else {
...
+    $vnetConfig.GWsubnetName = $GatewaySubnetName
+    $vnetConfig.RWsubnetName = $RemoteWorkstationSubnetName
+    if( $deployOverDC ) {
+        # Don't create new DC and vnets
+        # prompt for vnet name, gateway subnet name, remote workstatin subnet name, connection service subnet name
+        do {
+            if ( -not $vnetConfig.vnetName ) {
+                $vnetConfig.vnetName = Read-Host "Provide Connection Service Vnet Name"
+            }
+            if ( -not (Find-AzureRmResource -ResourceGroupNameEquals $rgName -ResourceType "Microsoft.Network/virtualNetworks" -ResourceNameEquals $vnetConfig.vnetName) ) {
```
bill-dall: I'm not sure we want to enforce where the vnet lives in resource group land - I figured we'd allow then to just submit the whole vnetID and it could be in any resource group. Also not sure if we're going to get into 'access rights' issues then about verifying that it exists. Hmm- not sure what the best plan is here.

Deploy-CAM.ps1
cptera: I was when I was stepping through the code, tenenatID was always null

Deploy-CAM.ps1
cptera: Yes, not neccessary, will remove

Deploy-CAM.ps1
cptera: My thinking (and I may be wrong) was that the only part of their infrastructure that we would need to hook into is the Vnet and subnets since the DC and storage accounts and whatever can be in separate resource groups or even on premise or whatever so it made sense to me to just assume the root group we want to overlay would be the one with the vnet since it's the only really mandatory thing. I can change it no problem, but I thought this was cleaner and easier since they don't need to provide that super long string.

Deploy-CAM.ps1
cptera: i jumbled the default names up when switching between branches, will change back to the correct originals

Deploy-CAM.ps1
bill-dall: Hmmm- which version? I am seeing both populated here. I think I'm on 5.0.1

bill-dall: @roy-satadru - do you recall which one was 'gone' in 5.0.0 - Tenant.Id or Tenant.TenantId ?

Deploy-CAM.ps1
```diff
@@ -1211,11 +1236,17 @@ function New-ConnectionServiceDeployment() {
     $artifactsLocation = $secret.SecretValueText
 
     Write-Host "Using service principal $client in tenant $tenant and subscription $subscriptionId"
-    New-AzureRmRoleAssignment `
-        -RoleDefinitionName Contributor `
-        -ResourceGroupName $csRGName `
-        -ServicePrincipalName $client `
-        -ErrorAction Stop | Out-Null
+    try {
+        New-AzureRmRoleAssignment `
```
bill-dall: I think best to change over to the 'check if needed and then assign' method (see other PR) rather than catch the one exception. I wonder what other exceptions we'll hit...

Deploy-CAM.ps1
roy-satadru: Sorry, I had missed this. Yes, I think Tenant.TenantId was removed from 5.0.0. So was Subscription.SubscriptionId and instead they only had the attributes Id. In 5.0.1 they added the longer attributes back.

Deploy-CAM.ps1
```diff
@@ -39,9 +39,29 @@ param(
...
+
+    [parameter(Mandatory=$false)]
+    [String]
+    $RemoteWorkstationSubnetName,    
+
     $camSaasUri = "https://cam-antar.teradici.com",
-    $CAMDeploymentTemplateURI = "https://raw.githubusercontent.com/teradici/deploy/bddc2b/azuredeploy.json",
-    $CAMDeploymentBlobSource = "https://teradeploy.blob.core.windows.net/bdstable",
+	$CAMDeploymentTemplateURI = "https://raw.githubusercontent.com/teradici/deploy/bd/azuredeploy.json",
+	$CAMDeploymentBlobSource = "https://teradeploy.blob.core.windows.net/bdbinaries",
```
bill-dall: We're now on bdstable...

Deploy-CAM.ps1
bill-dall: Okay, so we're guessing which one will cause the least pain going forward. Since we don't know I guess I'm fine with the change :)

Deploy-CAM.ps1
```diff
@@ -1731,9 +1786,79 @@ function Deploy-CAM() {
...
+                    Write-Host "Giving $client Contributor access to $rgn."
+                    New-AzureRmRoleAssignment `
+                        -RoleDefinitionName Contributor `
+                        -ResourceGroupName $rgn `
+                        -ServicePrincipalName $client `
+                        -ErrorAction Stop | Out-Null
+                }
+            }
+            
+            # Add Scopt to vNet if vNet alread exists and scope does not already exist
```
bill-dall: Scopt?

Deploy-CAM.ps1
```diff
@@ -1764,8 +1819,47 @@ function Deploy-CAM() {
                         -ErrorAction Stop | Out-Null
                 }
             }
+            
+            # Add Scopt to vNet if vNet alread exists and scope does not already exist
```
bill-dall: Scopt?

Deploy-CAM.ps1
```diff
@@ -2101,6 +2220,7 @@ if ($ResourceGroupName) {
         New-AzureRmResourceGroup -Name $ResourceGroupName -Location $location
     } 
     $rgMatch = Get-AzureRmResourceGroup -Name $ResourceGroupName
+    $rgName = $rgMatch.ResourceGroupName
```
bill-dall: Do we still need it? I thought I nuked the requirement for rgName (probably confusion because of base branch change.)

Deploy-CAM.ps1
```diff
@@ -2134,6 +2254,7 @@ else {
             }
             else {
                 $rgMatch = $resouceGroups[$rgIndex - 1]
+                $rgName = $rgMatch.ResourceGroupName
```
bill-dall: Ditto.

Deploy-CAM.ps1
cptera: ill double check, i added it back just to be safe

Deploy-CAM.ps1
```diff
@@ -39,6 +39,26 @@ param(
 	[String]
 	$AgentChannel = "stable",
 
+    [parameter(Mandatory=$false)]
```
bill-dall: Switch Parameter? https://powershell.org/forums/topic/understanding-switch-parameters/

Deploy-CAM.ps1
```diff
@@ -2233,7 +2354,92 @@ else {
...
+    $vnetConfig = @{}
+    $vnetConfig.vnetID = $vnetID
+    $vnetConfig.CSsubnetName = $CloudServiceSubnetName
+    $vnetConfig.GWsubnetName = $GatewaySubnetName
+    $vnetConfig.RWsubnetName = $RemoteWorkstationSubnetName
+    if( $deployOverDC ) {
+        # Don't create new DC and vnets
+        # prompt for vnet name, gateway subnet name, remote workstatin subnet name, connection service subnet name
+        do {
+            if ( -not $vnetConfig.vnetID ) {
```
bill-dall: I like that this can be prompted or done a a parameter. Nice!

Deploy-CAM.ps1
cptera: looks like we can nuke it easy, will do

Deploy-CAM.ps1
```diff
@@ -1131,235 +1151,325 @@ function New-ConnectionServiceDeployment() {
...
+                -Name "connectionServiceNumber" `
+                -SecretValue (ConvertTo-SecureString $connectionServiceNumber -AsPlainText -Force) `
+                -ErrorAction stop | Out-Null
+            
+            Write-Host "Checking available resource group for connection service number $connectionServiceNumber"
+
+            $csRGName = $RGName + "-CS" + $connectionServiceNumber
+            Set-AzureRMContext -Context $adminAzureContext | Out-Null
+            $rg = Get-AzureRmResourceGroup -ResourceGroupName $csRGName -ErrorAction SilentlyContinue
+            Add-AzureRmAccount `
```
bill-dall: I think this has to go after line 1277

bill-dall: Otherwise the SP may not have access.

Deploy-CAM.ps1
```diff
@@ -1131,235 +1151,325 @@ function New-ConnectionServiceDeployment() {
...
+                {
+                    # found the resource group was not empty - do the loop with an incremented number try to find a free name
+                    $csRGName = $null
+                }
+            }
+        }
+        
+        Write-Host "Using service principal $client in tenant $tenant and subscription $subscriptionId"
+        # Create Connection Service Resource Group if it doesn't exist
+        if (-not (Find-AzureRmResourceGroup | ?{$_.name -eq $csRGName}) ) {
```
bill-dall: This bit needs to be in the admin context, not the SP context.

Deploy-CAM.ps1
```diff
@@ -1131,235 +1151,325 @@ function New-ConnectionServiceDeployment() {
...
-        -ServicePrincipalName $client `
-        -ErrorAction Stop | Out-Null
+        if(-not $hasAccess) {
+            Write-Host "Giving $client Contributor access to $csRGName."
+            Set-AzureRMContext -Context $adminAzureContext | Out-Null
+            $rg = Get-AzureRmResourceGroup -ResourceGroupName $RGName -ErrorAction stop
+            $location = $rg.Location
+
+            Write-Host "Creating resource group $csRGName"
+            New-AzureRmResourceGroup -Name $csRGName -Location $location -ErrorAction stop | Out-Null
```
bill-dall: Did we just do this in line 1287?

Deploy-CAM.ps1
```diff
@@ -1629,13 +1746,12 @@ function Deploy-CAM() {
     $CAMConfig.parameters.AzureKeyVaultName = @{}
 	
     $CAMConfig.internal = @{}
-    $CAMConfig.internal.vnetName = "vnet-CloudAccessManager"
+    $CAMConfig.internal.vnetID = $vnetConfig.vnetID
+    $CAMConfig.internal.vnetName = $CAMConfig.internal.vnetID.split("/")[-1]
```
bill-dall: Below it's using split("/")[8] - should probably use the same criteria

Deploy-CAM.ps1
cptera: you're right

connection-service/new-admin-vm/Install-ConnectionServer.ps1
```diff
@@ -877,122 +877,116 @@ brokerIpaddress=$ipaddressString
...
+                        $content = @(
+                            '-----BEGIN CERTIFICATE-----'
+                            [System.Convert]::ToBase64String($caCert.RawData, 'InsertLineBreaks')
+                            '-----END CERTIFICATE-----'
+                        )
+                       
+                        $content | Out-File -FilePath "$env:systemdrive\$issuerCertFileName" -Encoding ascii
+                        $foundCert=$true
+                    } catch {
+                        throw "Failed to retrieve issuer certificate from $hostname`:$port because $_"
```
bill-dall: @cptera - I don't think we want to re-throw here until the retires are used up. (Or some other similar restructuring that makes sense :)

Deploy-CAM.ps1
```diff
@@ -2312,9 +2312,14 @@ else {
 
     $chosenSubscriptionNumber = 0 #invalid
     while ( -not (( $chosenSubscriptionNumber -ge 1) -and ( $chosenSubscriptionNumber -le $subscriptionsToDisplay.Length))) {
-        $chosenSubscriptionNumber = 
-        if (($chosenSubscriptionNumber = Read-Host "Please enter the Number of the subscription you would like to use or press enter to accept the current one [$currentSubscriptionNumber]") -eq '') `
-        {$currentSubscriptionNumber} else {$chosenSubscriptionNumber}
+        if( -not $ignorePrompts ) {
+            $chosenSubscriptionNumber =
```
cptera: This line won't work

bill-dall: Missing back-tick?

bill-dall: Actually somehow it did work. I don't know why. But yeah can add the back-tick.

Deploy-CAM.ps1
```diff
@@ -2333,7 +2333,18 @@ $rmContext = Set-AzureRmContext -SubscriptionId $subscriptionsToDisplay[$chosenS
...
 $selectedSubcriptionId = $subscriptionsToDisplay[$chosenSubscriptionIndex].SubscriptionId
 
-# Now we have the subscription set. Time to find the CAM root RG.
+# Now we have the subscription set. Ensure it has keyvault resource provider
+$keyVaultProviderExists = [bool](Get-AzureRmResourceProvider | Where-Object {$_.ProviderNamespace -eq "Microsoft.Keyvault"})
+
+if(-not $keyVaultProviderExists) {
+    Write-Host "Microsoft.Keyvault is not registered as a resource provider for this subscription."
+    Write-Host "Cloud Access Manager requires a key vault to operate."
+    $cancelDeployment = (Read-Host "Please hit enter to register Microsoft.Keyvault with subscription $($rmContext.Subscription.Id) or 'no' to cancel deployment") -like "*n*"
```
cptera: Should we add ignorePrompt  check to this?

bill-dall: Yes. I'll do that.

PR #63, opened 2017-12-06
Check if the SP needs a role assignment before attempting.

bill-dall:
This should work for the 'SP deploying itself' issue. @cptera -note the strange extra filters I had to put on to see if it had rights already. Seems like a loose pattern match that they have by default...

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -2142,24 +2142,25 @@ else {
             # entered a name. Let's see if it matches any resource groups first
             $rgMatch = $resouceGroups | Where-Object {$_.ResourceGroupName -eq $rgIdentifier}
             if ($rgMatch) {
-                $rgName = $rgMatch.ResourceGroupName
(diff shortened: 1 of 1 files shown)


Deploy-CAM.ps1
```diff
@@ -2111,24 +2142,25 @@ else {
...
             $rgMatch = $resouceGroups | Where-Object {$_.ResourceGroupName -eq $rgIdentifier}
             if ($rgMatch) {
-                $rgName = $rgMatch.ResourceGroupName
-                Write-Host ("Resource group `"$rgName`" already exists. The current one will be used.")
+                Write-Host ("Resource group `"$($rgMatch.ResourceGroupName)`" already exists. The current one will be used.")
                 $selectedRGName = $true
             }
             else {
                 # make a new resource group and on failure go back to RG selection.
-                $rgName = $rgIdentifier
```
bill-dall: The other way to solve the rgName issue is to get rid of the variable so we don't need to keep remembering to set it :) @cptera - I think this affects your PR as well.

PR #64, opened 2017-12-11
Better input validation

ivpetrov87:
Not sure if this catches every possible combinatin but according to  https://msdn.microsoft.com/en-us/library/ms853959.aspx only these are not valid \ / : * ? " < > |. We can expand on it further if we see something that is not handled

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -2267,8 +2267,8 @@ else {
             $domainName = Read-Host "Please enter new fully qualified domain name including a '.' such as example.com"
         }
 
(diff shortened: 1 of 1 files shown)


Deploy-CAM.ps1
```diff
@@ -2260,7 +2266,9 @@ else {
         if ( -not $domainName ) {
             $domainName = Read-Host "Please enter new fully qualified domain name including a '.' such as example.com"
         }
-        if ($domainName -notlike "*.*") {
+
+        # follwing https://msdn.microsoft.com/en-us/library/ms853959.aspx all but  \ / : * ? " < > | constitute valid netbios name
```
bill-dall: follwing

PR #66, opened 2017-12-12
Adding sumologic collector for linux agent and fix issues where idle …

jackyycheng:
…script hang.


remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -308,6 +314,31 @@ do
     fi
 done
 
+# Install and setup the Sumo Collector
+echo "-->Install SumoLogic collector"
+mkdir /tmp/sumo
```
bdyck: Rather than creating multiple temp directories, it might be better to create a temp directory once using `mktemp -d` and then use that as a working directory. It would also ensure that the temporary directory is uniquely named and secured (i.e. accessible only by the current user).

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -308,6 +314,31 @@ do
...
+wget "https://collectors.sumologic.com/rest/download/linux/64" -O /tmp/sumo/SumoCollector.sh && sudo chmod +x /tmp/sumo/SumoCollector.sh
+wget "$ARTIFACT_LOCATION/user.properties$SAAS_TOKEN" -O /tmp/sumo/user.properties
+wget "$ARTIFACT_LOCATION/sumo-agent-vm-linux.json$SAAS_TOKEN" -O /tmp/sumo/sumo-agent-vm-linux.json
+JSON_FILE=/tmp/sumo/sumo-agent-vm-linux.json
+echo "Attemtping to set sumo collector ID to: " "$COLLECTOR_ID"
+sed -i s/collectorID/"$COLLECTOR_ID"/ /tmp/sumo/user.properties
+sed -i 's|syncsourceFile|'$JSON_FILE'|' /tmp/sumo/user.properties
+
+sudo /tmp/sumo/SumoCollector.sh -q -varfile user.properties
```
bdyck: Are we assuming that this will always succeed? Do we care if it fails?

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -308,6 +314,31 @@ do
...
+service collector install
+service collector restart
+service collector status
+
+# Install idle shutdown script
+echo "-->Install idle shutdown"
+mkdir /tmp/idleShutdown
+wget "$ARTIFACT_LOCATION/Install-Idle-Shutdown.sh$SAAS_TOKEN" -O /tmp/idleShutdown/Install-Idle-Shutdown-raw.sh 
+awk '{ sub("\r$", ""); print }' /tmp/idleShutdown/Install-Idle-Shutdown-raw.sh > /tmp/idleShutdown/Install-Idle-Shutdown.sh && sudo chmod +x /tmp/idleShutdown/Install-Idle-Shutdown.sh
+sudo /tmp/idleShutdown/Install-Idle-Shutdown.sh -install
```
bdyck: Here again -- are we assuming that this will always succeed? If it does fail, do we care or need to take any additional steps?

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
jackyycheng: I would like to separate different component to different folder. It would be easier to change the code in the future

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
jackyycheng: I don't think this would impact user at all if it fails. Just harder to do debug.

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
jackyycheng: We wrote the idle shutdown script, I don't think it has any checking if it succeed or not.

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
bdyck: They can still be in subfolders -- they would just be inside the same top-level directory created by `mktemp`.

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
bdyck: If we don't really care if it succeeds or fails, then I think we should at least check for failure and ensure that we don't execute the `service` calls in that case.

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
bdyck: That's a bit troubling... :) In any case, since power management is an important feature in CAM, I think we should make an effort to at least detect if the Install-Idle-Shutdown.sh script fails, regardless of whether it does any error checking internally.

I'm not sure what the appropriate response is but my feeling is that a CAM user would want to know if they deployed an instance that was unable to automatically shut down when idle. Or maybe the appropriate response is to just fail the instance deployment entirely... ?

jackyycheng: I can create a story and see if we want to do it now?

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -68,6 +68,12 @@ PASSWORD="$6"
 GROUP="$7"
 # the eighth argument is the agent installer channel
 AGENT_CHANNEL="$8"
+# the nigth argument is the artifact location
```
bill-dall: ninth

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -68,6 +68,12 @@ PASSWORD="$6"
...
+ARTIFACT_LOCATION="$9"
+# the tenth argument is the sumo collector ID
+COLLECTOR_ID="${10}"
+# the eleventh argument is the SaaS Storage account token
```
bill-dall: SaaS or SaS? I think Satadru just renamed the rest of the system to SaS

remote-workstations/new-agent-vm/rhel-standard-agent.json
```diff
@@ -266,30 +267,7 @@
...
                 },
                 "protectedSettings": {
-                    "commandToExecute": "[concat('bash ', variables('scriptFileName'), ' \"', parameters('registrationCode'), '\" \"', parameters('agentType'), '\" \"', variables('vmSettings').vmNamePrefix, '\" \"', parameters('domainToJoin'), '\" \"', parameters('domainUsername'), '\" \"', parameters('domainPassword'), '\" \"', parameters('domainGroupToJoin'), '\" \"', parameters('AgentChannel'), '\"')]",
-                    "storageAccountName": "[parameters('userStorageAccountName')]",
-                    "storageAccountKey": "[parameters('userStorageAccountKey')]"
-                }
-            }
-        },
-        {
-            "type": "Microsoft.Compute/virtualMachines/extensions",
```
bill-dall: Getting rid of this was supposed to get rid of the hang...?

jackyycheng: Yes, it looks like it's the older extension hanging the script.

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
bill-dall: Yeah feels like a backlog item. We've already stretched this story a bit :)

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
jackyycheng: I'll create separate issue if we want to enhance this since this code is from CM deployment shell script

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
jackyycheng: I'll create separate issue if we want to enhance this since this code is from CM deployment shell script

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
bdyck: I think it's a fairly important one, though, because if the idle shutdown install script fails, it effectively breaks the power management feature. Since we're working on allowing custom images, it seems extra important to me that we make sure our install scripts are robust as we cannot assume they will be executing on a pristine, freshly-pressed OS. :)

PR #68, opened 2017-12-19
remove admin-rw storage account, data storage account

fwang-teradici:


remote-workstations/new-agent-vm/server2016-graphics-agent.json
```diff
@@ -226,7 +220,7 @@
                     "osDisk": {
                         "name": "osdisk",
                         "vhd": {
-                            "uri": "[concat('http://',parameters('storageAccountName'),'.blob.core.windows.net/vhds/',parameters('dnsLabelPrefix'), 'disk0.vhd')]"
```
bill-dall: I think we want the ability for the 'user' account (where binaries are stored) and the 'vhd' account to be separate, at least form the machine template standpoint. This will allow us to do multi-region deployments where the binaries are maybe only in one region but the VHD accounts are in different regions. I think I'd back out the changes in the machine templates and just point both parameters to the same account for now. that make sense?

root/azuredeploy.json
```diff
@@ -73,8 +73,7 @@
     }
   },
   "variables": {
-    "storageAccountName": "[concat(uniquestring(resourceGroup().id), 'adsa')]",
-    "templateStorageAccountName": "[concat(uniquestring(resourceGroup().id), 'data')]",
+    "storageAccountName": "[concat('adsa', uniquestring(resourceGroup().id))]",
```
bill-dall: Good idea to prepend the name. Maybe to keep the naming more consistent with the VM we're making it for, should we prepend 'vmdc' instead of 'adsa'?

fwang-teradici: I used adsa because it was used as subfix. and I thought it means for ad machine. Ya,  I agree vmdc is more suitable for this case.

remote-workstations/new-agent-vm/server2016-graphics-agent.json
fwang-teradici: then how about create storage account in *-RW group and pass to template even for the admin-rw and used to hold VHD

Deploy-CAM.ps1
```diff
@@ -331,7 +331,8 @@ function New-UserStorageAccount {
         $location
     )
 
-    $saName = -join ((97..122) | Get-Random -Count 18 | % {[char]$_})
+    $saName = -join ((97..122) | Get-Random -Count 16 | % {[char]$_})
+    $saName = 'cam0' + $saName
```
cptera: why cam0?

bill-dall: Needed some prefix that had only lower case letters and numbers... Nice restrictive naming!

PR #70, opened 2017-12-20
Check AzureRM version before starting

jackyycheng:


Deploy-CAM.ps1
```diff
@@ -2261,10 +2261,18 @@ function Deploy-CAM() {
     }
 }
 
+function Confirm-ModuleVersion {
+    # Check Azure RM version
+    if ( (Get-Module -ListAvailable -Name "AzureRM").Version.ToString() -lt "5.0.1") {
```
cptera: This sets the minimum supported version to 5.0.2 instead of 5.0.1

cptera: nevermind

Deploy-CAM.ps1
```diff
@@ -2261,10 +2261,18 @@ function Deploy-CAM() {
...
+        Write-Host ("AzureRM version must be equal or greater than 5.0.1")
```
cptera: This seems to do a character by character comparison so v"5.0.7" will be seen as a newer version than v"5.0.11"

cptera: also ("10.0.0" -lt "2.0.0") evaluates to True so I think a better way to do this is to split the version into major.minor.patch and then compare major minor and patch as ints instead of strings

PR #74, opened 2018-01-03
Remove unneeded parameters.

bill-dall:
Improve error handling around authentication race conditions.
Improve version checking handling to handle SDK clean install case.

Three things that I ran into yesterday...

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -2286,6 +2286,7 @@ function Confirm-ModuleVersion()
 
         if ( -not $AzureModule ) {
             # neither module found
+            Write-Host ("Please install the Azure Command Line tools for Powershell from Microsoft. The Azure and AzureRM modules must be present.")
             return $false
         }
         if ( [version]$AzureModule.Version.ToString() -lt [version]$MinAzureVersion) {
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -2284,11 +2284,13 @@ function Confirm-ModuleVersion()
         $MinAzureVersion="5.0.0"
         $AzureModule = Get-Module -ListAvailable -Name "Azure"
 
-        if ( $AzureModule ) {
-            if ( [version]$AzureModule.Version.ToString() -lt [version]$MinAzureVersion) {
-                Write-Host ("Azure module version must be equal or greater than " + $MinAzureVersion)
-                return $false
-            }
+        if ( -not $AzureModule ) {
+            # neither module found
+            return $false
+        }
+        if ( [version]$AzureModule.Version.ToString() -lt [version]$MinAzureVersion) {
+            Write-Host ("Azure module version must be equal or greater than " + $MinAzureVersion)
+            return $false
         }
     }
     return $true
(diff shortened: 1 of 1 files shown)


Deploy-CAM.ps1
```diff
@@ -4,9 +4,7 @@
 
 
 param(
-    $subscriptionId,
```
bill-dall: As @cptera pointed out, these two don't do anything...

Deploy-CAM.ps1
```diff
@@ -1499,11 +1497,12 @@ function New-ConnectionServiceDeployment() {
                 -Verbose
         }
         else {
-            for($idx = 10;$idx -gt 0;$idx--)
+            $maxRetries = 30
```
bill-dall: I hit this a lot yesterday, sometimes requiring 15 retries. I bumped the threshold to 30.

Deploy-CAM.ps1
```diff
@@ -1499,11 +1497,12 @@ function New-ConnectionServiceDeployment() {
...
+            for($idx = 0;$idx -lt $maxRetries;$idx++)
             {
                 try {
                     New-AzureRmResourceGroupDeployment `
-                        -DeploymentName "CS" `
+                        -DeploymentName "CS$connectionServiceNumber-$idx" `
```
bill-dall: If the deployment name is not unique then the deployment that was going to succeed will fail.

Deploy-CAM.ps1
```diff
@@ -2261,19 +2267,43 @@ function Deploy-CAM() {
     }
 }
 
-function Confirm-ModuleVersion {
+function Confirm-ModuleVersion()
```
bill-dall: Confirm-ModuleVersion was giving me errors on my laptop. The reason was or seems to be that AzureRM as a module is not installed with the newest SDK installer, just a new version of Azure module (and all the AzureRM sub-modules...) so this seems to be a valid configuration...

Deploy-CAM.ps1
```diff
@@ -2261,19 +2267,43 @@ function Deploy-CAM() {
...
+            Write-Host ("AzureRM module version must be equal or greater than " + $MinAzureRMVersion)
+            return $false
+        }
     }
+    else {
+        # the Azure SDK doesn't install 'AzureRM' as a base module any more, just Azure
+        $MinAzureVersion="5.0.0"
+        $AzureModule = Get-Module -ListAvailable -Name "Azure"
+
+        if ( $AzureModule ) {
```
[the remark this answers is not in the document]
bill-dall: Good point!

Deploy-CAM.ps1
```diff
@@ -2261,19 +2267,45 @@ function Deploy-CAM() {
...
+            return $false
+        }
+    }
+    else {
+        # the Azure SDK doesn't install 'AzureRM' as a base module any more, just Azure
+        $MinAzureVersion="5.0.0"
+        $AzureModule = Get-Module -ListAvailable -Name "Azure"
+
+        if ( -not $AzureModule ) {
+            # neither module found
```
jackyycheng: Would it be better to log something here so user know what to do?

PR #76, opened 2018-01-05
Passed owner upn to CAM Service

bnorouzizadeh:


Deploy-CAM.ps1
```diff
@@ -135,6 +180,48 @@ function Login-AzureRmAccountWithBetterReporting($Credential) {
...
+    $token.AccessToken
+}
+
+function Get-OwnerUpn() {
+    try {
+		$accessToken = Get-AzureRmCachedAccessToken
+		$decodedToken = Get-DecodedJWT `
+			-Token $accessToken `
+
+		$decodedToken.claims.upn
```
cptera: is this suppowed to get returned?

Deploy-CAM.ps1
```diff
@@ -135,6 +180,48 @@ function Login-AzureRmAccountWithBetterReporting($Credential) {
...
+        $azureRmProfile = [Microsoft.WindowsAzure.Commands.Common.AzureRmProfileProvider]::Instance.Profile
+        if(-not $azureRmProfile.Context.Account.Count) {
+            Write-Error "Ensure you have logged in before calling this function."
+        }
+    }
+    $currentAzureContext = Get-AzureRmContext
+    $profileClient = New-Object Microsoft.Azure.Commands.ResourceManager.Common.RMProfileClient($azureRmProfile)
+    Write-Debug ("Getting access token for tenant" + $currentAzureContext.Subscription.TenantId)
+    $token = $profileClient.AcquireAccessToken($currentAzureContext.Subscription.TenantId)
+    $token.AccessToken
```
cptera: return?

Deploy-CAM.ps1
```diff
@@ -135,6 +180,48 @@ function Login-AzureRmAccountWithBetterReporting($Credential) {
...
+}
+
+function Get-OwnerUpn() {
+    try {
+		$accessToken = Get-AzureRmCachedAccessToken
+		$decodedToken = Get-DecodedJWT `
+			-Token $accessToken `
```
cptera: I don't think you need that last `?

cptera:
could you also replace the tabs with 4 spaces just so we don't get weird indentations

Deploy-CAM.ps1
bill-dall: Yeah a return at the end of a PowerShell function is just 'style' so not really needed but my C++ is showing in my preference for having it :)

Deploy-CAM.ps1
```diff
@@ -135,6 +180,48 @@ function Login-AzureRmAccountWithBetterReporting($Credential) {
...
+
+		$decodedToken.claims.upn
+	}
+	catch {
+		$errorMessage = "An error occured while retrieving owner upn."
```
bill-dall: Have you tried this in Cloud Shell as well? I don't know if the token in Cloud Shell contains a UPN or a GUID?

Deploy-CAM.ps1
```diff
@@ -1601,6 +1697,8 @@ function New-CAMDeploymentRoot()
         -client $client `
         -key $key `
         -tenant $tenant `
+		-ownerTenant $ownerTenant `
```
bill-dall: I'd like to test this with my setup - in my case I sometimes use bdall@teradici.com to deploy to the primesoft.teradici.com subscription which has its own tenant, so the tenantID's might not match in that case. I don't know what the implication of that is...

Deploy-CAM.ps1
bnorouzizadeh: I tried this in PowerShell and not in Cloud Shell. However, the place to retrieve the access token is important because the context changes after calling Deploy-CAM and the token after that doesn't contain UPN anymore. That's why I am calling Get-OwnerUpn right before Deploy-CAM gets called.

Deploy-CAM.ps1
bnorouzizadeh: No need for return keyword.

Deploy-CAM.ps1
bill-dall: Okay - I think we need to ensure success in Cloud Shell before merging.

Deploy-CAM.ps1
```diff
@@ -202,6 +294,8 @@ public class TrustAllCertsPolicy : ICertificatePolicy {
                 username = $client
                 password = $key
                 tenantId = $tenant
+				ownerTenantId = $ownerTenant
```
cptera: What's the difference between tenantId and ownerTenantId?

Deploy-CAM.ps1
```diff
@@ -202,6 +294,8 @@ public class TrustAllCertsPolicy : ICertificatePolicy {
...
+                ownerUpn = $ownerUpn
```
cptera: I thought we already had an "email" field for registering users? Are we just not using it then?

bill-dall:
Thanks for testing with SP. Note you have a conflict with master now so will need to pull to get the merge to happen. So for clarification - if they deploy with an SP we just get an empty UPN?

I think we can merge as-is but going forward maybe should think about prompting or what Teradici wants to do with these cases of no UPN.

PR #78, opened 2018-01-08
Internal external

bill-dall:
Give the app gateway an internal IP in case external routing is not available (port 8443)
Rename some parameters for better clarity in the key vault
Make enableSecurityGateway switch for deployment of an overlay install or a connection service
Fix sumo paths in CM/SG machine

### DIFF
--- a/azuredeploy.json
+++ b/azuredeploy.json
@@ -31,13 +31,13 @@
     "localAdminUsername": {
       "type": "string",
       "metadata": {
-        "description": "The name of the local admin account for connection service VM's"
+        "description": "The name of the local admin account for connection service VM's and the domain controller, if created"
       }
     },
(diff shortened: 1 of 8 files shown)


azuredeploy.json
```diff
@@ -31,13 +31,13 @@
     "localAdminUsername": {
       "type": "string",
       "metadata": {
-        "description": "The name of the local admin account for connection service VM's"
+        "description": "The name of the local admin account for connection service VM's and the domain controller, if created"
```
cptera: Should VM be Remote Workstations?

bill-dall: In this case not - it's the local admin password for the security gateways, connection server, and domain controller (if created).

PR #80, opened 2018-01-10
Connection server does not join domain.

bill-dall:
CS get an auto-generated static IP through a two-step ARM template process.
CM uses IP based routing instead of DNS (since the name will be less deterministic after its not domain joined).
CS no-longer needs to remote-powershell to DC to get LDAPS certificate. (DC now always makes it before saying its done.)

Bonus: Fix up the password checking for the domain account in Deploy-CAM.ps1

This seemed like a good intermediate step to merge since it is working without any noticeable side-effects, and has a useful fix or two.

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -2678,12 +2678,6 @@ else {
             $domainAdminCredential = Get-Credential -Message $domainAdminMessage
             $confirmedPassword = Read-Host -AsSecureString "Please re-enter the password"
 
-            if (-not ($domainAdminCredential.UserName -imatch '\w+') -Or ($domainAdminCredential.Username.Length -gt 20)) {
-                Write-Host "Please enter a valid username. It can only contain letters and numbers and cannot be longer than 20 characters."
-                $domainAdminCredential = $null
-                continue
-            }
-
             # Need plaintext password to check if same
             $clearPassword = ConvertTo-Plaintext $confirmedPassword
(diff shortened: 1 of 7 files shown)


Deploy-CAM.ps1
```diff
@@ -2694,10 +2688,17 @@ else {
...
-		
-        if ($domainAdminCredential.GetNetworkCredential().Password.Length -lt 12) {
+        if ((-not ($domainAdminCredential.UserName -imatch '\w+')) -or ($domainAdminCredential.Username.Length -gt 20)) {
+            Write-Host "Please enter a valid username. It can only contain letters and numbers and cannot be longer than 20 characters."
+            $domainAdminCredential = $null
+            continue
+        }
+
+        if ((-not $deployOverDC) `
+            -and ( $domainAdminCredential.GetNetworkCredential().Password.Length -lt 12 ) ) {
```
cptera: Should we add in the other checks here as well? @bdyck ran into that issue this week where he had "admin" in the Password or something and it errored out much later

bill-dall: Good point - let's make a ticket for those. I think that was to do with the username and not the password, but I think we should also add the 'username should not be in password' check and maybe a couple other things. The reason this is in this branch is because this fix isn't a 'try it again' type thing - the service account would need to be modified if their service account doesn't match our password policy, which is silly and inconvenient. Luckily one of Jeremy's accounts met our threshold.

bill-dall:
Note the 'don't actually join the domain' step was pretty small from where this PR was at so I added that in as well. Exciting!

PR #81, opened 2018-01-10
Fixed Install Logic for Sumo-Logic:

cptera:
* Wait for service to be installed instead of waiting for registry key


cptera:
https://github.com/teradici/deploy/pull/81/files?w=1 to ignore whitespace changes

PR #82, opened 2018-01-11
RADIUS MFA enablement 

roy-satadru:


Deploy-CAM.ps1
```diff
@@ -1813,6 +1860,9 @@ function Deploy-CAM() {
 
         [parameter(Mandatory = $true)]
         $vnetConfig,
+
+        [parameter(Mandatory=$false)]
```
bill-dall: I think it's mandatory or you will have de-referencing null object issues in the function.

Deploy-CAM.ps1
```diff
@@ -2349,6 +2414,38 @@ function Deploy-CAM() {
 				},
 				"secretName": "cloudAccessRegistrationCode"
 			}
+		},
```
bill-dall: Seems to be some tabs versus spaces thing going on in this section? I think @cptera changed the whole file recently to 4-spaces indents.

Deploy-CAM.ps1
```diff
@@ -64,9 +64,24 @@ param(
 
     [switch]$ignorePrompts,
 
+    [parameter(Mandatory = $false)]
+    $enableRadiusMfa=$null,
```
bill-dall: I think it's $null if not specified (like the ones below) but no harm explicitly setting to $null I guess.

Deploy-CAM.ps1
```diff
@@ -64,9 +64,24 @@ param(
...
+    $radiusServerPort,
+
+    [parameter(Mandatory=$false)]
+    [SecureString]
+    $radiusSharedSecret,
+
     $camSaasUri = "https://cam.teradici.com",
 	$CAMDeploymentTemplateURI = "https://raw.githubusercontent.com/teradici/deploy/master/azuredeploy.json",
-    $binaryLocation = "https://teradeploy.blob.core.windows.net/binaries",
+    $binaryLocation = "https://teradeploy.blob.core.windows.net/binaries/",
```
bill-dall: What symptom drove this?

Deploy-CAM.ps1
```diff
@@ -2812,6 +2909,56 @@ else {
         }
     } while ( -not $domainAdminCredential )
 
+    $radiusConfig = @{
+        enableRadiusMfa = $enableRadiusMfa
+        radiusServerHost = $radiusServerHost
+        radiusServerPort = $radiusServerPort 
+        radiusSharedSecret = $radiusSharedSecret
+    }
+    if ( ($enableRadiusMfa -eq $null) -or $enableRadiusMfa ) {
```
bill-dall: Why enter the prompt area if it's already set? I think the if's here can probably be simplified.

Deploy-CAM.ps1
bill-dall: But if you don't set to null then you can make it a [bool] type which would be cleaner.

Deploy-CAM.ps1
```diff
@@ -2812,6 +2909,56 @@ else {
...
+            } while (-not $radiusConfig.radiusServerPort )
+
+            if (-not $radiusConfig.radiusSharedSecret ) {
+                $radiusConfig.radiusSharedSecret = Read-Host -AsSecureString "Please enter your Radius Server's Shared Secret"
+            }
+        }
+        $radiusConfig.enableRadiusMfa = $enableRadiusMfa
+    } 
+    
+    if ( -not $enableRadiusMfa) {
```
bill-dall: rather than another if I think this could be an 'else'

Deploy-CAM.ps1
```diff
@@ -2812,6 +2909,56 @@ else {
...
+                        $radiusConfig.radiusServerPort = [int](Read-Host  "Please enter your Radius Server's Listening port")
+                    } catch {
+                        $radiusConfig.radiusServerPort = $null
+                        Write-Host "Selected port is not an Integer"
+                    }
+                    if ( ($radiusConfig.radiusServerPort -le 0) -or ($radiusConfig.radiusServerPort -gt 65535) ) {
+                        Write-Host "Selected port is invalid. Should be between 1 and 65535."
+                        $radiusConfig.radiusServerPort = $null
+                    } else {
+                        $radiusConfig.radiusServerPort = $radiusConfig.radiusServerPort
```
bill-dall: Is the else needed here? Seems a no-op.

Deploy-CAM.ps1
```diff
@@ -2812,6 +2909,56 @@ else {
...
+        } elseif ( $enableRadiusMfa -eq $null -and $ignorePrompts ) {
+            $enableRadiusMfa = $false
+        }
+
+        if ($enableRadiusMfa) {
+            if (-not $radiusConfig.radiusServerHost ) {
+                $radiusConfig.radiusServerHost = Read-Host "Please enter your Radius Server's Hostname or IP"
+            }
+
+            do {
```
bill-dall: Good to have the checking and retry loop :)

Deploy-CAM.ps1
```diff
@@ -2812,6 +2909,56 @@ else {
...
+                if (-not $radiusConfig.radiusServerPort ) {
+                    try {
+                        $radiusConfig.radiusServerPort = [int](Read-Host  "Please enter your Radius Server's Listening port")
+                    } catch {
+                        $radiusConfig.radiusServerPort = $null
+                        Write-Host "Selected port is not an Integer"
+                    }
+                    if ( ($radiusConfig.radiusServerPort -le 0) -or ($radiusConfig.radiusServerPort -gt 65535) ) {
```
bill-dall: If you pop this 'if' up one level it will also check passed parameters, not just command-line inputs (see the other do loops in the code.)

azuredeploy.json
```diff
@@ -163,14 +163,42 @@
       "metadata": {
         "description": "The URL of resources, such as templates, scripts, and DSC modules, that the ARM deployment template depends on"
       },
-      "defaultValue": "https://raw.githubusercontent.com/teradici/deploy/master"
```
bill-dall: Need to back this out before commit. (And I don't think we need to have a default here at all actually since deploy-CAM.ps1 always passes a value.)

connection-service/new-admin-vm/new-admin-vm.json
```diff
@@ -50,7 +50,35 @@
                 "description": "Auto-generated token to access _artifactsLocation"
             },
             "defaultValue": ""
-        }
+        },
+        "enableRadiusMfa" : {
+            "type": "string",
+            "metadata": {
+              "description": "Enable RAIDUS based MFA?"
```
bill-dall: RADIUS

Deploy-CAM.ps1
```diff
@@ -2812,6 +2909,56 @@ else {
...
+                        $radiusConfig.radiusServerPort = $radiusConfig.radiusServerPort
+                    }
+                }
+            } while (-not $radiusConfig.radiusServerPort )
+
+            if (-not $radiusConfig.radiusSharedSecret ) {
+                $radiusConfig.radiusSharedSecret = Read-Host -AsSecureString "Please enter your Radius Server's Shared Secret"
+            }
+        }
+        $radiusConfig.enableRadiusMfa = $enableRadiusMfa
```
bill-dall: I'm really curious what string this is set to. It seems to me we're not constraining it so it's whatever the person passed. I think we're looking for the JSON standard string for 'true'?

Deploy-CAM.ps1
```diff
@@ -2812,6 +2909,56 @@ else {
...
+    } 
+    
+    if ( -not $enableRadiusMfa) {
+        # Placeholder value for the radius secret and port is required in order to create KeyVault entry
+        $radiusConfig.radiusSharedSecret = ConvertTo-SecureString "radiusSecret" -AsPlainText -Force
+        $radiusConfig.radiusServerPort = ConvertTo-SecureString 0 -AsPlainText -Force
+        
+        $radiusConfig.radiusServerHost = ConvertTo-SecureString "radiusServer" -AsPlainText -Force
+    }
```
bill-dall: Should we be explicitly setting $radiusConfig.enableRadiusMfa to 'false' here in this block?

connection-service/new-admin-vm/Install-ConnectionServer.ps1
```diff
@@ -874,6 +891,23 @@ brokerProductVersion=1.0
 brokerIpaddress=$ipaddressString
 brokerLocale=en_US
 "@
+              
+                $isMfa = $using:enableRadiusMfa
+#                Write-Host "MFA setting is $isMfa"
+#stick in RADIUS MFA related attributes if RADIUS MFA is turned on
+                if($isMfa -eq "True") {
```
bill-dall: -eq isn't case sensitive, which I think is a good thing in this case.

connection-service/new-admin-vm/Install-ConnectionServer.ps1
```diff
@@ -874,6 +891,23 @@ brokerProductVersion=1.0
...
+                    $localRadiusHost = $using:radiusServerHost
```
bill-dall: inside the string (below) you can just use $using:radiusServerHost embedded in the string and avoid the local's. (I didn't know that when I wrote a bunch of this code.)

Deploy-CAM.ps1
bill-dall: Oh I see we're doing that now :)

Deploy-CAM.ps1
```diff
@@ -2824,9 +2824,12 @@ else {
...
     if ( -not $enableRadiusMfa) {
         # Placeholder value for the radius secret and port is required in order to create KeyVault entry
+        $radiusConfig.enableRadiusMfa = $false
         $radiusConfig.radiusServerSharedSecret = ConvertTo-SecureString "radiusSecret" -AsPlainText -Force
-        $radiusConfig.radiusServerPort = ConvertTo-SecureString 0 -AsPlainText -Force
-        $radiusConfig.radiusServerHost = ConvertTo-SecureString "radiusServer" -AsPlainText -Force
+        $radiusConfig.radiusServerPort = 0
+        $radiusConfig.radiusServerHost = "radiusServer"
+    } else {
+        $radiusConfig.enableRadiusMfa = $true
```
bill-dall: Should this move up to the previous 'if' block?

Deploy-CAM.ps1
cptera: This was my hacky way of getting three states for this:
1. If True, skip prompts and enable Radius settings
2. If False, skip prompt and disable Radius settings
3. If Null, go through prompts and set as desired by user
Bool limits me to three states, but I can change it and force the prompt unless ignorePrompts is set if you want

Deploy-CAM.ps1
cptera: Not needed, fixing

Deploy-CAM.ps1
cptera: You're right, easier/safer if I just make this mandatory

Deploy-CAM.ps1
cptera: For each option it prompts only if it's not set

Deploy-CAM.ps1
cptera: I can probably change it now that I think about it some more...

connection-service/new-admin-vm/Install-ConnectionServer.ps1
roy-satadru: Yes, I could have used "true" as well but as long as I din't use -ceq I figured we would be ok.

connection-service/new-admin-vm/new-admin-vm.json
roy-satadru: Will fix.

connection-service/new-admin-vm/Install-ConnectionServer.ps1
roy-satadru: I didn't know that. Fixing it.

Deploy-CAM.ps1
bill-dall: Yeah I think setting to bool will still allow the three state bit you want (which I agree with) but just get the whole 'did I type the text correctly' part out of the equation.

azuredeploy.json
roy-satadru: Will fix.

Deploy-CAM.ps1
bill-dall: Okay - I thought I saw a logic error here but no it looks fine.

Deploy-CAM.ps1
roy-satadru: FWIW, I did print out the type of enableRadiusMfa with a call to getType() and it did print bool but that was after it was set to $true/$false.

PR #85, opened 2018-01-19
Linux graphics agent

fwang-teradici:


remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -1,122 +1,117 @@
...
+{
+    echo 'blacklist nouveau' | sudo tee -a /etc/modprobe.d/nouveau.conf
+    
+    echo 'blacklist lbm-nouveau' | sudo tee -a /etc/modprobe.d/nouveau.conf
+}
+
+# need to reboot after install Linux Integration Services for Hyper-V
+install_lis()
+{    
+    local LIS_FILE="lis-rpms-4.2.3-4.tar.gz"
```
cptera: Are we going to have to periodically update this?

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -243,131 +238,326 @@ active_d = ldap_lib(args.address, args.user_name, args.password, args.domain)
...
 
+    sudo python $file_path -d "$DOMAIN_NAME" -a "$DC_ADDRESS" -u "$USERNAME" -p "$PASSWORD" -c "$VM_NAME" -g "$GROUP"
+}
+
+install_gui() 
+{
+    # Make sure Linux OS is up to date
+    echo "--> Updating Linux OS to latest"
+
+    sudo yum -y update  # --exclude=WALinuxAgent
```
cptera: Does this work? We needed to exclude WALinuxAgent before

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -243,131 +238,326 @@ active_d = ldap_lib(args.address, args.user_name, args.password, args.domain)
...
+    INST_LAST_STEP=$(tail -1 $INST_LOG_FILE) 
+fi
+
+if [ "$INST_LAST_STEP" == "initial" ]
+then
+    echo "start installing pcoip $AGENT_TYPE agent" | tee $INST_LOG_FILE
+
+    echo "step1 starting" | tee -a $INST_LOG_FILE
+
+    sudo yum install wget
```
cptera: Don't you need a -y here?

fwang-teradici: it seems no need here

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -1,122 +1,117 @@
 #!/bin/bash
 
-install_driver()
+if [ $# -eq 1 ]
```
bdyck: We should use `[[ ]]` for tests, as it's safer than `[ ]` and the shebang line says this is running in Bash. There are several places where this should be changed...

(`[ ]` is for `sh` and can have different semantics)

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -1,122 +1,117 @@
...
+
+  sudo yum -y install kernel-devel
+
+  sudo rpm -Uvh --quiet https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+
+  sudo yum -y install dkms
+}
+
+# need to reboot after install
+disable_nouveal()
```
bdyck: I think there is a typo here... This should be `disable_nouveau()`.

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -243,131 +238,326 @@ active_d = ldap_lib(args.address, args.user_name, args.password, args.domain)
...
+        else
+            INST_LAST_STEP="step6 failure: $EXIT_CODE"
+        fi
+
+        echo "$INST_LAST_STEP" | tee -a $INST_LOG_FILE
+    fi
+fi
+
+#remove job
+sudo crontab -r
```
cptera: Does running this of a crontab job was never created do anything? It looks like it's possible to get here without creating cronjob

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -1,122 +1,117 @@
...
+    
+    echo 'blacklist lbm-nouveau' | sudo tee -a /etc/modprobe.d/nouveau.conf
+}
+
+# need to reboot after install Linux Integration Services for Hyper-V
+install_lis()
+{    
+    local LIS_FILE="lis-rpms-4.2.3-4.tar.gz"
+    
+    wget --retry-connrefused --tries=3 --waitretry=5  http://download.microsoft.com/download/6/8/F/68FE11B8-FAA4-4F8D-8C7D-74DA7F2CFC8C/$LIS_FILE
```
bdyck: I see a couple of things here...

* The URL should be enclosed in double quotes as it's a string
* The URL is using HTTP -- we should use HTTPS

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: yes for standard agent, there is no cronjob.  and remove operation is harmless operation here

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
bdyck: Good question... the version number should probably be a script variable.

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: if fact, the can move up into if statement

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -243,131 +238,326 @@ active_d = ldap_lib(args.address, args.user_name, args.password, args.domain)
...
+install_gui() 
+{
+    # Make sure Linux OS is up to date
+    echo "--> Updating Linux OS to latest"
+
+    sudo yum -y update  # --exclude=WALinuxAgent
+
+    # Install Desktop
+    echo "-->Install desktop"	
+    #sudo yum -y groupinstall 'X Window System' 'GNOME'
```
bdyck: There are several commented-out commands or debug statements in this function that I think we could just remove for the sake of clarity. :)

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -243,131 +238,326 @@ active_d = ldap_lib(args.address, args.user_name, args.password, args.domain)
...
+            ;;
+        "dev")
+            agent_repo_url="https://downloads.teradici.com/rhel/pcoip-dev.repo"
+            ;;   
+        *)
+            agent_repo_url="https://downloads.teradici.com/rhel/pcoip.repo"
+            ;;       
+    esac
+    
+    sudo wget --retry-connrefused --tries=3 --waitretry=5 -O /etc/yum.repos.d/pcoip.repo $agent_repo_url
```
bdyck: The `$agent_repo_url` variable should be in double quotes.

bill-dall:
With all these reboots, what's the experience in the CAM management interface? I'm guessing it gets to 'running' and 'deployment complete' fairly quickly but won't be actually connectable for some time after that. (I don't think this is gating for this story/pull request but we should probably know what to expect from a user perspective.)

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -243,131 +238,326 @@ active_d = ldap_lib(args.address, args.user_name, args.password, args.domain)
...
-            echo "failed to register pcoip agent license."
-            # let's define exit code 102 for this case
-            exit 102
+        INST_LAST_STEP="step3 failure: $EXIT_CODE"
+    fi
+
+    echo "$INST_LAST_STEP" | tee -a $INST_LOG_FILE
+fi
+
+if [ $AGENT_TYPE == $AGENT_TYPE_GRAPHICS ]
```
bdyck: String comparisons should be in double quotes.

fwang-teradici:
@bill-dall  You are right. After the first reboot, deploy is completed from "azure view". But the installation is not completed in fact, for graphic agent, it has to reboots another twice

bill-dall:
Okay, thanks @fwang-teradici

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: I will change

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: will change

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: good catch. it is from original file

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: using double quote

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: using double quotes

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: it works

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: at this time no, We control which version to install like windows

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: it is possible.  Do we need to exclude forever to solve the issue?

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: version number as variable did no provide any benefit at here

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
fwang-teradici: let's keep those comments here

PR #86, opened 2018-01-23
Configure auto shutdown idle timer

bdyck:
This change allows users to configure both the idle timer for the auto shutdown feature (i.e. how long a machine must be idle before automatic shutdown is triggered) or whether the feature is enabled at all. These settings can be configured in the ARM templates with the `autoShutdownIdleTime` parameter (idle time in minutes) and the `enableAutoShutdown` parameter (boolean).

(Note that the idle shutdown service is always installed; the `enableAutoShutdown` flag simply controls whether or not it's enabled on the remote workstation.)

The default idle time is definitely open for debate... :)

### DIFF
--- a/remote-workstations/new-agent-vm/Install-Idle-Shutdown.sh
+++ b/remote-workstations/new-agent-vm/Install-Idle-Shutdown.sh
@@ -250,7 +250,7 @@ OnUnitActiveSec=15min
 Unit=${SERVICE}
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=timers.target
 EOF
     cat <<EOF> ${SERVICE_CONFIG}
 [Service]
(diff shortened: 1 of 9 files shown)


remote-workstations/new-agent-vm/Install-Idle-Shutdown.sh
```diff
@@ -13,6 +19,16 @@ SERVICE_CONFIG="${SERVICE_CONFIG_PATH}/CAMIdleShutdown.conf"
 
 MONITOR_SCRIPT=/opt/Teradici_CAM_idle_shutdown.py
 
+# Idle time in minutes before auto-shutdown should be engaged
+IDLE_TIMER=240
```
jackyycheng: Any reason why change for 60 to 240 mins? That's 4 hours....

remote-workstations/new-agent-vm/Install-PCoIPAgent.ps1
```diff
@@ -26,6 +26,9 @@ Configuration InstallPCoIPAgent
         [Parameter(Mandatory=$false)]
         [bool]$enableAutoShutdown,
 
+		[Parameter(Mandatory=$false)]
```
jackyycheng: spacing?

remote-workstations/new-agent-vm/Install-Idle-Shutdown.sh
bdyck: @bill-dall and I talked about this a bit and felt that 60 minutes seemed too short...

jackyycheng: If will only shut down if there is no PCoIP session. If there is a PCoIP session, even if user is doing nothing, it won't get shut down.

PR #87, opened 2018-01-23
in method New-ConnectionServiceDeployment and Deploy-CAM, using try b…

fwang-teradici:
…lock to cover more code, so we can restore the context


cptera:
Instead of moving the try-catch-finally to cover a bit more code, why not make a main function and do a try-catch-finally around that?

fwang-teradici:
@cptera 1) there are two place to call New-ConnectionServiceDeployment 2) you only need to cover what you need to cover (it run a little fast). 3) the try / catch /finally is already there.

PR #92, opened 2018-01-30
Dc run on 2012r2

fwang-teradici:


Deploy-CAM.ps1
```diff
@@ -79,6 +79,11 @@ param(
     [SecureString]
     $radiusSharedSecret,
 
+    [parameter(Mandatory=$false)]
+    [ValidateSet("Windows Server 2016","Windows Server 2012")]
```
bill-dall: I would not say 2012. I would say 2012R2. Historically, it was very important which one of the two people were using and I feel it can be confusing if we're specifying the wrong one.

fwang-teradici: OK, change to 2012R2

PR #93, opened 2018-02-01
Use managed disks

cptera:


Deploy-CAM.ps1
```diff
@@ -2519,6 +2526,59 @@ function Confirm-ModuleVersion()
     return $true
 }
 
+# Create a custom role for CAM with necessary permissions
+# Currently it is the based off of the Contributor Role excpept we remove Storage Access and VM Disk Blob Read Access
```
bill-dall: I think the implication here is that the account cannot access any storage account (unless it gets a token from the keyvault...) but really that means that we ONLY support managed storage now since it can no longer create the disks for unmanaged storage?

Deploy-CAM.ps1
```diff
@@ -2519,6 +2526,59 @@ function Confirm-ModuleVersion()
...
+        $camCustomRoleDefinition.AssignableScopes.Clear()
+        $camCustomRoleDefinition.AssignableScopes.Add("/subscriptions/$subscriptionId")
+
+        New-AzureRmRoleDefinition -Role $camCustomRoleDefinition | Out-Null
+        $camCustomRoleDefinition = Get-AzureRmRoleDefinition "Cloud Access Manager"
+    }
+
+    # Actions to remove
+    $requiredNotActions = @(
+        # Remove ability to perform any storage accoutn actions
```
bill-dall: accoutn

Deploy-CAM.ps1
```diff
@@ -2519,6 +2526,59 @@ function Confirm-ModuleVersion()
...
+        "Microsoft.Storage/storageAccounts/*",
+        # Remove ability to get SAS URI of VM Image for Blob access
```
bill-dall: Are they still called blobs here? disk access?

Deploy-CAM.ps1
```diff
@@ -2519,6 +2526,59 @@ function Confirm-ModuleVersion()
...
+        "Microsoft.Compute/disks/endGetAccess/action"
+    )
+    # Add Not Actions required to be disabled
+    foreach ( $notAction in $requiredNotActions) {
+        if ( -not $camCustomRoleDefinition.NotActions.Contains($notAction)) {
+            $camCustomRoleDefinition.NotActions.Add($notAction)
+        }
+    }
+
+    # Add Subscription to Assignable Scopes if not already there
```
bill-dall: What does this do?

connection-service/new-vm-join-domain/new-vm-join-domain.json
```diff
@@ -88,15 +82,14 @@
...
         "windowsOSVersion": "2016-Datacenter",
-        "apiVersion": "2015-06-15",
         "nicName": "[concat(parameters('dnsLabelPrefix'),'Nic')]",
         "updateipTemplateUri": "[concat(parameters('_artifactsLocation'), '/update-nic.json', parameters('_artifactsLocationSasToken'))]"
     },
     "resources": [
         {
             "type": "Microsoft.Network/networkInterfaces",
             "name": "[variables('nicName')]",
-            "apiVersion": "[variables('apiVersion')]",
```
bill-dall: Yeah I agree it seems better to have the individual API dates we want than some global variable that always seems wrong...

connection-service/new-vm-join-domain/new-vm-join-domain.json
```diff
@@ -133,19 +126,13 @@
...
-                        "vhd": {
-                            "uri": "[concat('http://',parameters('storageAccountName'),'.blob.core.windows.net/vhds/',parameters('dnsLabelPrefix'), 'disk0.vhd')]"
-                        },
+                        "name": "[concat(parameters('dnsLabelPrefix'), '-os')]",
                         "caching": "ReadWrite",
                         "createOption": "FromImage"
                     },
                     "dataDisks": [
                         {
-                            "name": "myvm-data-disk1",
```
bill-dall: I'm sure this is from me eon's ago but I was never fully certain why we need the extra disks.

Deploy-CAM.ps1
cptera: Set's the subscriptions this can run on. By default it was "/" meaning everything, but I didnt' see any harm in limiting it down to the CAM subscription only

Deploy-CAM.ps1
cptera: That's what the documentation called it, "Blob access to the disk"

connection-service/new-vm-join-domain/new-vm-join-domain.json
bill-dall: Maybe the default OS disk size is too small?

Deploy-CAM.ps1
cptera: Yeah, we can't do anything with storage accounts after this change including making VHDs. Now that I think about it, this will probably make our VHD template stuff break...

connection-service/new-vm-join-domain/new-vm-join-domain.json
cptera: Yeah, I didn't really see the point either, but didn't bother changing it. Default size is 128 GB.

cptera: According to https://docs.microsoft.com/en-us/azure/virtual-machines/windows/about-disks-and-vhds , maximum size for OS disk is 2 TB so maybe that's why

Deploy-CAM.ps1
bill-dall: Good point on the template stuff - yeah we can't break that.

Deploy-CAM.ps1
bill-dall: Okay sounds good. Whatever works :)

Deploy-CAM.ps1
bill-dall: Okay. Sounds good.

connection-service/new-vm-join-domain/new-vm-join-domain.json
bill-dall: 2 TB is pretty big :) Okay with a 128 GB default, and the increased pricing of managed storage, I think we can nuke the second drives.

Deploy-CAM.ps1
cptera: Yeah, VHD templates are broken :(

bill-dall: I think it's okay to have blob read access in the resource groups by default as long as we don't have disk read access in the resource groups.

Deploy-CAM.ps1
```diff
@@ -2519,6 +2519,57 @@ function Confirm-ModuleVersion()
...
+# Currently it is the based off of the Contributor Role excpept we remove Storage Access and VM Disk Blob Read Access
+# Use 'Get-AzureRmProviderOperation *' to get a list of Azure Operations and their details
+# See https://docs.microsoft.com/en-us/azure/active-directory/role-based-access-built-in-roles for details on Azure Built in Roles
+function Get-CAMRoleDefinition() {
+    param(
+        [parameter(Mandatory = $true)]
+        [String]$subscriptionId
+    )
+
+    $camCustomRoleDefinition = Get-AzureRmRoleDefinition "Cloud Access Manager"
```
bill-dall: What happens if the account doesn't have the correct access rights for the 'roledefinition' API's? Uncaught exception? Other badness?

cptera: Yeah, it throws an permissions error.

bill-dall: Okay, as long as the error is pretty clear and things exit, sounds good.

Deploy-CAM.ps1
```diff
@@ -2519,6 +2519,54 @@ function Confirm-ModuleVersion()
...
+# Currently it is the based off of the Contributor Role excpept we remove Storage Access and VM Disk Blob Read Access
+# Use 'Get-AzureRmProviderOperation *' to get a list of Azure Operations and their details
+# See https://docs.microsoft.com/en-us/azure/active-directory/role-based-access-built-in-roles for details on Azure Built in Roles
+function Get-CAMRoleDefinition() {
+    param(
+        [parameter(Mandatory = $false)]
+        [String]$subscriptionId
+    )
+
+    $camCustomRoleDefinition = Get-AzureRmRoleDefinition "Cloud Access Manager"
```
bill-dall: There are 6 occurrences of the role string "Cloud Access manager" in the code, which could be single sourced but your call.

cptera: good point

PR #94, opened 2018-02-02
Deploy continue after join group failure

fwang-teradici:


remote-workstations/new-agent-vm/Install-PCoIPAgent.ps1
```diff
@@ -519,40 +519,62 @@ Configuration InstallPCoIPAgent
...
+                            Remove-PSSession $psSession
+                        }
                     }
-                    Remove-PSSession $psSession
                 }
                     
                 #make placeholder file so this is only run once
                 New-Item "$using:agentInstallerDLDirectory\domainGroupJoinFile.txt" -type file
+
+                if ($exception -ne $null) {
```
bill-dall: Does this mean the DSC configuration doesn't complete if the domain join group failed?

fwang-teradici: No. DSC still complete after join group failed because it still generate the flag file. It catch the exception and write the exception information to file

bill-dall: Ah okay. And you're just writing the exception information to the file for debugging later?

fwang-teradici: yes, you are right. Only exception information to the file for debugging. if joining group successful, then the flag file is empty as before.

PR #96, opened 2018-02-07
- Make warning more visible

jackyycheng:
- Fixed domain name checking
- Add checking for banned username
- Fixed on not prompting username again if it is invalid password
- Add checking for password complexity

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -2583,6 +2583,13 @@ function Get-CAMRoleDefinition() {
     return $camCustomRoleDefinition
 }
 
+function Write-Host-Warning() {
+    param(
+        $message
+    )
+    Write-Host ("`n$message") -ForegroundColor Red -BackgroundColor Yellow
+}
+
 ##############################################
 ############# Script starts here #############
 ##############################################
@@ -2602,7 +2609,7 @@ $subscriptionsToDisplay = $subscriptions | Where-Object { $_.State -eq 'Enabled'
 
 $chosenSubscriptionIndex = $null
 if ($subscriptionsToDisplay.Length -lt 1) {
-    Write-Host ("Account " + $rmContext.Account.Id + " has access to no enabled subscriptions. Exiting.") -ForegroundColor Red -BackgroundColor Yellow
+    Write-Host-Warning -message "Account " + $rmContext.Account.Id + " has access to no enabled subscriptions. Exiting."
     exit
 }
 
@@ -2711,7 +2718,7 @@ else {
         $rgIdentifier = Read-Host "Resource group"
 
         if (!$rgIdentifier) {
-            Write-Host ("`nValue not provided.") -ForegroundColor Red -BackgroundColor Yellow
+            Write-Host-Warning -message "Value not provided."
             continue       
         }
 
@@ -2722,7 +2729,7 @@ else {
             $rgArrayLength = $resouceGroups.Length
             if ( -not (( $rgIndex -ge 1) -and ( $rgIndex -le $rgArrayLength))) {
                 #invalid range 
-                Write-Host "`nPlease enter a range between 1 and $rgArrayLength or the name of a new resource group." -ForegroundColor Red -BackgroundColor Yellow
+                Write-Host-Warning -message "Please enter a range between 1 and $rgArrayLength or the name of a new resource group."
             }
             else {
                 $rgMatch = $resouceGroups[$rgIndex - 1]
@@ -2863,7 +2870,7 @@ else {
                 -ResourceType "Microsoft.Network/virtualNetworks" `
                 -ResourceNameEquals $vnetName)) ) {
                     # Does not exist
-                    Write-Host ("`n{0} not found" -f ($vnetConfig.vnetID)) -ForegroundColor Red -BackgroundColor Yellow
+                    Write-Host-Warning -message "$($vnetConfig.vnetID) not found"
                     $vnetConfig.vnetID = $null
             }
         } while (-not $vnetConfig.vnetID)
@@ -2877,7 +2884,7 @@ else {
             }
             if ( -not ($vnet.Subnets | ?{$_.Name -eq $vnetConfig.CSsubnetName}) ) {
                 # Does not exist
-                Write-Host ("`n{0} not found in root resource group VNet {1}" -f ($vnetConfig.CSsubnetName,$vnet.Name)) -ForegroundColor Red -BackgroundColor Yellow
+                Write-Host-Warning -message "$($vnetConfig.CSsubnetName) not found in root resource group VNet $($vnet.Name)"
                 $vnetConfig.CSsubnetName = $null
             }
         } while (-not $vnetConfig.CSsubnetName)
@@ -2889,7 +2896,7 @@ else {
             }
             if ( -not ($vnet.Subnets | ?{$_.Name -eq $vnetConfig.GWsubnetName}) ) {
                 # Does not exist
-                Write-Host ("`n{0} not found in root resource group VNet {1}" -f ($vnetConfig.GWsubnetName,$vnet.Name)) -ForegroundColor Red -BackgroundColor Yellow
+                Write-Host-Warning -message "$($vnetConfig.GWsubnetName) not found in root resource group VNet $($vnet.Name)"
                 $vnetConfig.GWsubnetName = $null
             }
         } while (-not $vnetConfig.GWsubnetName)
@@ -2901,7 +2908,7 @@ else {
             }
             if ( -not ($vnet.Subnets | ?{$_.Name -eq $vnetConfig.RWsubnetName}) ) {
                 # Does not exist
-                Write-Host ("`n{0} not found in root resource group VNet {1}" -f ($vnetConfig.RWsubnetName,$vnet.Name)) -ForegroundColor Red -BackgroundColor Yellow
+                Write-Host-Warning -message "$($vnetConfig.RWsubnetName) not found in root resource group VNet $($vnet.Name)"
                 $vnetConfig.RWsubnetName = $null
             }
         } while (-not $vnetConfig.RWsubnetName)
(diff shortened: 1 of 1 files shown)


Deploy-CAM.ps1
```diff
@@ -2924,58 +2924,101 @@ else {
...
+                $domainNameMessage = "`nPlease enter the fully qualified domain name of the domain to connect to. The name must include a '.' such as example.com"
             }
             Write-Host $domainNameMessage
             $domainName = Read-Host "Domain name"
         }
 
-        # https://social.technet.microsoft.com/Forums/scriptcenter/en-US/db2d8388-f2c2-4f67-9f84-c17b060504e1/regex-for-computer-fqdn?forum=winserverpowershell
-        if (-not $($domainName -imatch '(?=^.{1,254}$)(^(?:(?!\d+\.|-)[a-zA-Z0-9_\-]{1,63}(?<!-)\.?)+(?:[a-zA-Z]{2,})$)')) {
-            Write-Host "The domain name must include two or more components separated by a '.'"
+        if (-not $($domainName -match '^[a-zA-Z0-9]{1,61}[a-zA-Z0-9](?:\.[a-zA-Z]{2,})+$')) {
```
bdyck: This definitely looks more straightforward than the previous regex but I have some concerns here:

- even though the `match` operator is case-insensitive, we should be explicit and use the `imatch` operator; this will also remove the need to account for uppercase characters in the regex
- `0-9` can be replaced with the digit class `\d`
- a hyphen (`-`) is a valid character in a domain name, but this regex doesn't allow it

Deploy-CAM.ps1
```diff
@@ -2924,58 +2924,101 @@ else {
...
-
-            # Need plaintext password to check if same
-            $clearPassword = ConvertTo-Plaintext $confirmedPassword
-            if (-not ($domainAdminCredential.GetNetworkCredential().Password -ceq $clearPassword)) {
-                # don't match- try again.
-                Write-Host "The entered passwords do not match."
-                $domainAdminCredential = $null
+        # only check if it is not deployOverDC
+        if (-not $deployOverDC) {
+            if ((-not ($username -match '^[A-Za-z0-9]+(?:[_-][A-Za-z0-9]+)*$')) -or ($username.Length -gt 20)) {
```
bdyck: I think some of the same comments apply to this regex:

- use `imatch` and remove uppercase character ranges
- should use `\d` for digits

Deploy-CAM.ps1
```diff
@@ -2924,58 +2924,101 @@ else {
...
+
+        if ( -not $password ) {
+            $psw = Read-Host -AsSecureString "Please enter the password"
+            $password = ConvertTo-Plaintext $psw
+            $prompted = 1
+        }
+
+        # Don't check password if deploying over DC since in that case it's the DC's password complexity rules.
+        if (-not $deployOverDC ) {
+            if (-not ($password -match '^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[$@$!%*?&])[A-Za-z\d$@$!%*?&]{12,}')) {
```
bdyck: I suspect that attempting to enforce password composition rules in a regex is an unwinnable fight... it will certainly be an unreadable one. :) I tried this regex against several strings that would be perfectly fine passwords elsewhere (generated via password generators and phrases including whitespace) but are rejected by this regex.

If we are going to enforce composition requirements, I think we'd be better off writing a method to handle it than attempting to use a regex.

Deploy-CAM.ps1
```diff
@@ -2602,7 +2602,7 @@ $subscriptionsToDisplay = $subscriptions | Where-Object { $_.State -eq 'Enabled'
 
 $chosenSubscriptionIndex = $null
 if ($subscriptionsToDisplay.Length -lt 1) {
-    Write-Host ("Account " + $rmContext.Account.Id + " has access to no enabled subscriptions. Exiting.")
+    Write-Host ("Account " + $rmContext.Account.Id + " has access to no enabled subscriptions. Exiting.") -ForegroundColor Red -BackgroundColor Yellow
```
bdyck: I think this particular combination of `Write-Host` colors could be extracted as a method, which would make the intent clear as well as de-clutter the rest of the script a bit.

Deploy-CAM.ps1
```diff
@@ -2924,58 +2924,101 @@ else {
...
         if ( -not $domainName ) {
             if( -not $deployOverDC ) {
-                $domainNameMessage = "Please enter new fully qualified domain name of the domain which will be created, including a '.' such as example.com"
+                $domainNameMessage = "`nPlease enter new fully qualified domain name of the domain which will be created, including a '.' such as example.com"
             }
             else {
-                $domainNameMessage = "Please enter the fully qualified domain name of the domain to connect to. The name must include a '.' such as example.com"
+                $domainNameMessage = "`nPlease enter the fully qualified domain name of the domain to connect to. The name must include a '.' such as example.com"
             }
             Write-Host $domainNameMessage
```
bill-dall: Add colors here?

Deploy-CAM.ps1
bill-dall: This regex is so complex it does hurt my brain... @ivpetrov87 found the original one from some online thing. Does the new one work with 3-level domains such as blah.blah.com ? I don't know what my pereference is here but the 'minimal change' part of me says that maybe what we want is the original regex plus a search for 'there must be a dot somewhere' to force a two-level domain.

Deploy-CAM.ps1
```diff
@@ -2924,58 +2924,101 @@ else {
...
+        if (-not $($domainName -match '^[a-zA-Z0-9]{1,61}[a-zA-Z0-9](?:\.[a-zA-Z]{2,})+$')) {
+            Write-Host "`nInvalid Domain name. Please see https://support.microsoft.com/en-ca/help/909264/naming-conventions-in-active-directory-for-computers-domains-sites-and for valid domain names." -ForegroundColor Red -BackgroundColor Yellow
             $domainName = $null
         }
+
     } while (-not $domainName)
 
+    # Username
+    $username = $null
+    $firstRun = 1
```
bill-dall: Is furstrun used in the user section?

Deploy-CAM.ps1
```diff
@@ -2924,58 +2924,101 @@ else {
...
     do {
-        if ( -not $domainAdminCredential ) {
+        if ($domainAdminCredential -and $firstRun) {
```
bill-dall: Why not pull this out of the do loop - the $domainAdminCredential.UserName just becomes a 'seed' for username.

Deploy-CAM.ps1
```diff
@@ -2924,58 +2924,101 @@ else {
...
-            $domainAdminCredential = Get-Credential -Message $domainAdminMessage
-            $confirmedPassword = Read-Host -AsSecureString "Please re-enter the password"
-
-            # Need plaintext password to check if same
-            $clearPassword = ConvertTo-Plaintext $confirmedPassword
-            if (-not ($domainAdminCredential.GetNetworkCredential().Password -ceq $clearPassword)) {
-                # don't match- try again.
-                Write-Host "The entered passwords do not match."
-                $domainAdminCredential = $null
+        # only check if it is not deployOverDC
```
bill-dall: good point!

Deploy-CAM.ps1
bill-dall: And just for interest- how is this different than imatch '\w+'?

Deploy-CAM.ps1
```diff
@@ -2924,58 +2924,101 @@ else {
...
 
-        if ((-not ($domainAdminCredential.UserName -imatch '\w+')) -or ($domainAdminCredential.Username.Length -gt 20)) {
-            Write-Host "Please enter a valid username. It can only contain letters and numbers and cannot be longer than 20 characters."
-            $domainAdminCredential = $null
-            continue
+    } while ( -not $username )
+
+    # Password
+    $password = $null
+    $firstRun = 1
```
bill-dall: $true?

Deploy-CAM.ps1
```diff
@@ -2924,58 +2924,101 @@ else {
...
+            }
         }
 
-        if ((-not ($domainAdminCredential.UserName -imatch '\w+')) -or ($domainAdminCredential.Username.Length -gt 20)) {
-            Write-Host "Please enter a valid username. It can only contain letters and numbers and cannot be longer than 20 characters."
-            $domainAdminCredential = $null
-            continue
+    } while ( -not $username )
+
+    # Password
```
bill-dall: This feels like it could be a lot simpler if we set both $password and $confirmedPassword to  $domainAdminCredential.GetNetworkCredential().Password outside of the do loop and then only prompt if not $null (and set both to $null on failure) then don't need the state variables.

Deploy-CAM.ps1
jackyycheng: Yes, it is used in both user and password session.

Deploy-CAM.ps1
jackyycheng: Made a function

Deploy-CAM.ps1
jackyycheng: humm...I could not directly set variable to $domainAdminCredential since it is System.Management.Automation.PSCredential object. That's why at line 3021 I create a new credential object

Deploy-CAM.ps1
jackyycheng: humm...I could not directly set variable to $domainAdminCredential since it is System.Management.Automation.PSCredential object. That's why at line 3021 I create a new credential object

Deploy-CAM.ps1
jackyycheng: I will remove the checking for password complexity to only check for 12 characters

Deploy-CAM.ps1
bdyck: @bill-dall Good catch. The rules for usernames are less restrictive so I think `\w+` would do fine. :)

Deploy-CAM.ps1
jackyycheng: Add another checking for dot only and use original regex

Deploy-CAM.ps1
jackyycheng: \w+ does not check for -

jackyycheng: Also, it seems you can't start or end with - or _

ivpetrov87: - should not match with \w. IS - accepted in usernames?

Deploy-CAM.ps1
bill-dall: Sorry, confused - you can set $username = $domainAdminCredential.UserName inside the do loop but not outside? (I know that's not what you said but I'm just a bit confused since it does seem that things can be set anywhere at any time, and you can create and extract credentials like you said.

Deploy-CAM.ps1
bill-dall: Yeah looks better with the function thanks!

Deploy-CAM.ps1
bill-dall: I'm fine with whichever regex works here. As long as it's tested.

Deploy-CAM.ps1
```diff
@@ -2602,7 +2610,7 @@ $subscriptionsToDisplay = $subscriptions | Where-Object { $_.State -eq 'Enabled'
 
 $chosenSubscriptionIndex = $null
 if ($subscriptionsToDisplay.Length -lt 1) {
-    Write-Host ("Account " + $rmContext.Account.Id + " has access to no enabled subscriptions. Exiting.")
+    Write-Host-Warning -message "Account " + $rmContext.Account.Id + " has access to no enabled subscriptions. Exiting."
```
cptera: I don't think you need the '-message'

PR #97, opened 2018-02-07
Expose idle shutdown time as script parameter

bdyck:
FYI: the min and max range come from the values accepted by the Windows agent's idle shutdown service.

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -84,6 +84,11 @@ param(
     [String]
     $domainControllerOsType = "Windows Server 2016",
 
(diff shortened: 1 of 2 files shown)


bill-dall:
@bdyck, @cptera , now won't automation have to wait 240 minutes?

bdyck:
@bill-dall No, we'll update the automation to use this parameter and then it will take whatever that value is.

bill-dall:
@bdyck - But not for admin-rw, then?

bdyck:
@bill-dall It will also pick up that value.

PR #102, opened 2018-02-15
Select vnets from a list, as well as subnets.

bill-dall:
Remove confirmation on vnet access rights - but make a warning string.
Trim all appropriate input since... that's the right thing to do for UX.

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -2043,6 +2043,7 @@ function Deploy-CAM() {
                 Write-Host "The Cloud Access Manager deployment script was not passed service principal credentials. It will attempt to create a service principal."
                 $requestSPGeneration = Read-Host `
                     "Please hit enter to continue or 'no' to manually enter service principal credentials from a pre-made service principal"
+                $requestSPGeneration = $requestSPGeneration.Trim()
             } else {
                 $requestSPGeneration=""
             }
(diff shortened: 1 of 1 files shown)


Deploy-CAM.ps1
```diff
@@ -3015,7 +3076,7 @@ else {
     }
     do {
         if ( -not $password ) {
-            $psw = Read-Host -AsSecureString "Please enter the password"
+            $psw = (Read-Host -AsSecureString "Please enter the password").Trim()
```
jackyycheng: Should we trim password? I thought spaces are allowed for password. I'm not sure if you can have leading or ending with spaces in password tho.

Deploy-CAM.ps1
```diff
@@ -3029,7 +3090,7 @@ else {
         }
 
         if ($psw) {
-            $confirmedPassword = Read-Host -AsSecureString "Please re-enter the password"
+            $confirmedPassword = (Read-Host -AsSecureString "Please re-enter the password").Trim()
```
jackyycheng: If we don't trim password, we should remove trim here too.

Deploy-CAM.ps1
bill-dall: Good point! And probably the RADIUS shared secret as well.

PR #108, opened 2018-02-23
Updated NVidia Windows Driver:

cptera:
* Updated to GRID 5.3 Driver (386.09)


cptera:
NOTE: This actually upgraded to GRID 5.2, not 5.3 like the comment says

PR #109, opened 2018-02-23
Fix invalid location issue

jackyycheng:


Deploy-CAM.ps1
```diff
@@ -2830,10 +2830,17 @@ else {
                 $inputRgName = $rgIdentifier
                 $newRgResult = $null
 
-                Write-Host("Available Azure Locations")
-                Write-Host (Get-AzureRMLocation | Select-Object -Property Location, DisplayName | Format-Table | Out-String )
-
-                $newRGLocation = (Read-Host "`nPlease enter resource group location").Trim()
+                $azureLocation = Get-AzureRMLocation
+                $locations = @(($azureLocation | Select-Object Location).location)
```
bill-dall: Actually - now that I think of it the previous implementation accepted both 'eastus' or 'east us' which was nice (the API accepts both columns). I think this implementation does not? If so I'd add the other column into the array as well.

PR #111, opened 2018-02-26
Improve user prompting for CAM deployment

fwang-teradici:


Deploy-CAM.ps1
```diff
@@ -2667,7 +2690,7 @@ function Write-Host-Warning() {
     param(
         $message
     )
-    Write-Host ("`n$message") -ForegroundColor Red -BackgroundColor Yellow
+    Write-Host ("`n$message") -ForegroundColor Red
```
jackyycheng: We all agree just to show Red Text?

fwang-teradici: Backgroud Color is what it is. It is more readable

PR #114, opened 2018-03-06
Modify sumologic names for deployment

fwang-teradici:


fwang-teradici:
1) remote workstation does not need this information for it gets from registrationCode
2) in Deploy-CAM.ps1, no new parameter added, licenseInstanceId  is calculated from registration code
3) only connection service template add parameter to accept licenseInstanceId

PR #115, opened 2018-03-08
Slightly improved prints.

bill-dall:


Deploy-CAM.ps1
```diff
@@ -2574,6 +2574,11 @@ function Confirm-ModuleVersion()
     return $true
 }
 
+function Get-CAMRoleDefinitionName() {
```
bill-dall: I moved this to a separate function since I was planning on using it from a few locations but then ended up with one location, but I thought in the end it's nicer to not be part of the larger function, below.

PR #118, opened 2018-03-09
remove public IP from DC

fwang-teradici:


bill-dall:
Microsoft has lots of instructions on how to add IP addresses - let's leave that to Michael and me and move on.

(Though there should still be a story created for Michael to update the doc.)

Thanks,

Bill

PR #119, opened 2018-03-13
Internal network

cptera:


Deploy-CAM.ps1
```diff
@@ -2709,6 +2661,155 @@ function Write-Host-Warning() {
...
+        $changeRadiusSettings = (confirmDialog ([String]::Format("Radius Multi-Factor Authentication is currently {0}. Do you want to change Multi-Factor Authentication settings?", $currentRadiusSetting))) -eq 'y'
+    } else {
+        # Check if any settings are different
+        $changeRadiusSettings = ($enableRadiusMfa -notmatch $isRadiusMfaEnabled) -or `
+                                ($radiusServerHost -and ($radiusServerHost -notmatch $currentRadiusHost)) -or `
+                                ($radiusServerPort -and ($radiusServerPort -notmatch $currentRadiusPort)) -or `
+                                ($radiusSharedSecret)
+    }
+
+    # Update Radius Settings appropriatley
```
bill-dall: appropriately

Deploy-CAM.ps1
```diff
@@ -2709,6 +2661,155 @@ function Write-Host-Warning() {
     Write-Host ("`n$message") -ForegroundColor Red
 }
 
+# Prompt update Radius Settings in the Keyvault
```
bill-dall: Prompt for and update... ?

Deploy-CAM.ps1
```diff
@@ -1503,6 +1502,9 @@ function New-ConnectionServiceDeployment() {
             -TenantId $tenantId `
             -ErrorAction Stop | Out-Null
 
+        Set-RadiusSettings `
```
bill-dall: Is this in the admin context or the SP context?

Deploy-CAM.ps1
```diff
@@ -2709,6 +2661,155 @@ function Write-Host-Warning() {
...
+    $currentRadiusPort = ([string]$currentRadiusPort.SecretValueText)
+
+    $currentRadiusSecret = Get-AzureKeyVaultSecret `
+        -VaultName $VaultName `
+        -Name "radiusSharedSecret" `
+        -ErrorAction stop
+    $currentRadiusSecret = ([string]$currentRadiusSecret.SecretValueText)
+
+    # Prompt for RADIUS configuration if RADIUS has not been already explicitly been disabled
+    if ($ignorePrompts -and ($enableRadiusMfa -eq $null)) {
```
bill-dall: Is enableRadiusMfa a global variable?

Deploy-CAM.ps1
```diff
@@ -2709,6 +2661,155 @@ function Write-Host-Warning() {
...
+        $enableRadiusMfa = $isRadiusMfaEnabled
+    }
+    if ($enableRadiusMfa -eq $null) {
+        $currentRadiusSetting = "disabled"
+        if ($isRadiusMfaEnabled) {
+            $currentRadiusSetting = "enabled"
+        }
+        $changeRadiusSettings = (confirmDialog ([String]::Format("Radius Multi-Factor Authentication is currently {0}. Do you want to change Multi-Factor Authentication settings?", $currentRadiusSetting))) -eq 'y'
```
bill-dall: Can't you just put $currentRadiusSetting directly in the dual-quotes quoted string without the format function?

Deploy-CAM.ps1
```diff
@@ -2709,6 +2661,155 @@ function Write-Host-Warning() {
...
+    } else {
+        # Check if any settings are different
+        $changeRadiusSettings = ($enableRadiusMfa -notmatch $isRadiusMfaEnabled) -or `
+                                ($radiusServerHost -and ($radiusServerHost -notmatch $currentRadiusHost)) -or `
+                                ($radiusServerPort -and ($radiusServerPort -notmatch $currentRadiusPort)) -or `
+                                ($radiusSharedSecret)
```
bill-dall: It's always changed if there is a secret present?

Deploy-CAM.ps1
bill-dall: This setting it to stuff and then checking for $null seems a bit dodgy - can we use more Boolean like logic or is there three states we're worrying about?

Deploy-CAM.ps1
```diff
@@ -2709,6 +2661,155 @@ function Write-Host-Warning() {
...
+                    do {
+                        $radiusConfig.radiusServerHost = (Read-Host "Enter your RADIUS Server's Hostname or IP").Trim()
+                    } while (-not $radiusConfig.radiusServerHost)
+                }
+            } elseif ($radiusServerHost) {
+                $radiusConfig.radiusServerHost = $radiusServerHost
+            }
+
+            if ((-not $radiusServerPort) -and (-not $ignorePrompts)) {
+                if((confirmDialog "Do you want to change your RADIUS Server Port?") -eq 'y') {
```
bill-dall: Tell them what the current one is?

Deploy-CAM.ps1
```diff
@@ -2709,6 +2661,155 @@ function Write-Host-Warning() {
...
+                if((confirmDialog "Radius Server Host is currently $currentRadiusHost. Do you want to change your RADIUS Server Host?") -eq 'y') {
+                    do {
+                        $radiusConfig.radiusServerHost = (Read-Host "Enter your RADIUS Server's Hostname or IP").Trim()
+                    } while (-not $radiusConfig.radiusServerHost)
+                }
+            } elseif ($radiusServerHost) {
+                $radiusConfig.radiusServerHost = $radiusServerHost
+            }
+
+            if ((-not $radiusServerPort) -and (-not $ignorePrompts)) {
```
bill-dall: This whole block seems to be avoidable with $ignorePrompts - can we deal with that earlier and avoid checking?

Deploy-CAM.ps1
```diff
@@ -2709,6 +2661,155 @@ function Write-Host-Warning() {
...
+                $radiusConfig.radiusSharedSecret = $radiusSharedSecret
+            }
+        }
+
+        # Store required keys as CamConfig Object
+        $camConfig = @{
+            parameters=@{}
+        }
+        foreach ($key in $radiusConfig.keys) {
+            if ($key -eq "radiusSharedSecret") {
```
bill-dall: Can we just check type?

connection-service/azuredeploy.json
```diff
@@ -308,6 +278,12 @@
           },
           "CSUniqueSuffix": {
             "value": "[parameters('CSUniqueSuffix')]"
+          },
+          "enableExternalAccess": {
+            "value": "[bool(parameters('enableExternalAccess'))]"
+          },
+          "enableRadiusMfa": {
```
bill-dall: We could just make this part of CAMDeploymentParameters but I'm not sure it's worth it...

connection-service/new-app-gateway/azuredeploy.json
```diff
@@ -65,6 +65,19 @@
         "description": "Application Gateway Name including the unique CS suffix"
       }
     },
+    "enableExternalAccess": {
+      "type": "bool",
+      "metadata": {
+          "description": "'true' if a Public IP address should be created, otherwise 'false'."
+      }
+    },
+    "enableRadiusMfa": {
```
bill-dall: In this file I don't think it has anything to do with MFA, it's more about enabling external access to the management interface? I'd suggest renaming.

connection-service/new-app-gateway/azuredeploy.json
```diff
@@ -238,68 +351,27 @@
         ],
         "urlPathMaps": [
           {
-            "name": "urlPathMap1",
+            "name": "pcoipBrokerProtocolUserInterfacePath",
```
bill-dall: broker protocol?

cptera: The URL path does both PBP if it starts with /pcoip-broker/ and otherwise redirects to the UI

connection-service/new-app-gateway/azuredeploy.json
cptera: good point

Deploy-CAM.ps1
cptera: Looks like SP context

Deploy-CAM.ps1
cptera: Yeah it's global and yeah it's three states (enabled, disable, unset)

Deploy-CAM.ps1
cptera: You're right, don't know why I did this

Deploy-CAM.ps1
cptera: Yeah, forgot to add that in

Deploy-CAM.ps1
cptera: Hmm, you're probably right

Deploy-CAM.ps1
cptera: I'm not sure what you mean

Deploy-CAM.ps1
cptera: fixed

Deploy-CAM.ps1
cptera: We can, but it looks like there's more code duplication and more lines if we do due to verifying the port and having to to duplicate the code for when the values are passed in

Deploy-CAM.ps1
bill-dall: Okay good. Thanks.

Deploy-CAM.ps1
bill-dall: Can we do this without a global variable? Three states is of course fine if there are three states, but looking at the code it seemed to be 'check something for $null that may or may not exist' which is easy to get wrong in PowerShell and cause weird bugs. I've done that a few times.

Deploy-CAM.ps1
bill-dall: I see. That makes more sense. I can re-review thinking more about 'when it's set or not.'

Deploy-CAM.ps1
bill-dall: I was just thinking - For each key if it's not a secure string make it a secure string. Don't check based on data structure key names. Then this code doesn't really care about the data structure.

connection-service/new-app-gateway/azuredeploy.json
bill-dall: Right - so one path map. Got it.

Deploy-CAM.ps1
bill-dall: Duh - that's set right above :)

Deploy-CAM.ps1
```diff
@@ -2709,6 +2661,155 @@ function Write-Host-Warning() {
     Write-Host ("`n$message") -ForegroundColor Red
 }
 
+# Prompt update Radius Settings in the Keyvault
+function Set-RadiusSettings() {
+    Param(
+        [string]$VaultName
+    )
+    # Check current MFA settings
+    $isRadiusMfaEnabled = Get-AzureKeyVaultSecret `
```
bill-dall: Okay, this is probably worth a bit of thought/discussion, but some of the complexity is that we're kind of keeping this key vault secret as a cached value which we sometimes want and sometimes don't. Maybe we should just always be asking.

Deploy-CAM.ps1
cptera: you're right, thats better

Deploy-CAM.ps1
cptera: fixed

Deploy-CAM.ps1
cptera: sure, no more globals

Deploy-CAM.ps1
```diff
@@ -2695,23 +2723,17 @@ function Set-RadiusSettings() {
...
-        # Load provided Radius Configuration Parameters
+    # Check if any settings are different
+    $changeRadiusSettings = `
+        (($enableRadiusMfa -ne $null)   -and ($enableRadiusMfa -notmatch $isRadiusMfaEnabled)) -or `
+        ($radiusServerHost              -and ($radiusServerHost -notmatch $currentRadiusHost)) -or `
+        ($radiusServerPort              -and ($radiusServerPort -notmatch $currentRadiusPort)) -or `
+        ($radiusSharedSecret)
+
+    # Update RADIUS Settings appropriately
+    if ( $changeRadiusSettings -or (-not $ignorePrompts) ) {
```
bill-dall: Right okay so this is saying (since real people don't use ignorePrompts) - if I'm a real person, use this code, if I'm automation only use this code if I'm passing values.

Deploy-CAM.ps1
```diff
@@ -2695,23 +2723,17 @@ function Set-RadiusSettings() {
...
 
-    # Update Radius Settings appropriatley
-    if ( $changeRadiusSettings ) {
-        # Load provided Radius Configuration Parameters
+    # Check if any settings are different
+    $changeRadiusSettings = `
+        (($enableRadiusMfa -ne $null)   -and ($enableRadiusMfa -notmatch $isRadiusMfaEnabled)) -or `
+        ($radiusServerHost              -and ($radiusServerHost -notmatch $currentRadiusHost)) -or `
+        ($radiusServerPort              -and ($radiusServerPort -notmatch $currentRadiusPort)) -or `
+        ($radiusSharedSecret)
```
bill-dall: So we could just compare this easily with a .SecretValueText suffix?

connection-service/new-app-gateway/azuredeploy.json
```diff
@@ -71,11 +71,11 @@
           "description": "'true' if a Public IP address should be created, otherwise 'false'."
       }
     },
-    "enableRadiusMfa": {
+    "disableExternalUserInterface": {
       "type": "bool",
       "defaultValue": false,
       "metadata": {
-        "description": "Indicates weather or not MFA is enabled"
+        "description": "Indicates weather or not to make UI extnerally accessible"
```
bill-dall: whether

bill-dall: "If true, disables external access to the management interface." ?

Deploy-CAM.ps1
cptera: Correct, automation should always go to the else block (assuming you're passing values)

Deploy-CAM.ps1
cptera: We do have the Password in the keyvault as a string, but we'd have to decode the one that a user would pass in. It's doable in a few lines of code to decode it, but I'm not sure if that's something we want to do.

connection-service/new-app-gateway/azuredeploy.json
cptera: fixed

Deploy-CAM.ps1
```diff
@@ -2739,7 +2765,7 @@ function Set-RadiusSettings() {
             }
 
             if ((-not $radiusServerPort) -and (-not $ignorePrompts)) {
-                if((confirmDialog "Do you want to change your RADIUS Server Port?") -eq 'y') {
+                if((confirmDialog "RADIUS Server Port is currently $currentRadiusPort. Do you want to change your RADIUS Server Port?") -eq 'y') {
                     do {
                         try {
                             $radiusConfig.radiusServerPort = [int](Read-Host  "Enter your RADIUS Server's Listening port")
```
bill-dall: Don't you get a weird error thrown if you just do the int cast this way - I had to do something like 'try int' elsewhere to get it to interpret it properly.

bill-dall: Ahh there it is: [int]::TryParse(

Deploy-CAM.ps1
bill-dall: Okay I see what you mean. I still have this nagging feeling there are simpler ways here but maybe not simple enough to matter.

Deploy-CAM.ps1
bill-dall: yeah okay let's not bother. I agree.

PR #120, opened 2018-03-15
Disable MFA for internal:

cptera:
* Internal Access with MFA is disabled

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -1321,11 +1321,11 @@ function New-ConnectionServiceDeployment() {
         $spCredential,
         $keyVault,
         $testDeployment,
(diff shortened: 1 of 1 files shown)


Deploy-CAM.ps1
```diff
@@ -2775,6 +2779,11 @@ function Set-RadiusSettings() {
         }
         $radiusConfig.enableRadiusMfa = $enableRadiusMfa
 
+        if ( $radiusConfig.enableRadiusMfa -and (-not $enableExternalAccess)) {
```
bill-dall: That right - I said it wrong in the email. The configuration that has the security issue is when external access is disabled and MFA is enabled. This feels a bit weird to deny it in this way but I think we can live with it for a while.

PR #122, opened 2018-03-26
* External Port should always be port 443

cptera:
Untested


cptera:
Bills changes look good to me

PR #123, opened 2018-04-04
Improve error messaging if AzureRM provides an error ID. (This happen…

bill-dall:
…s in the common quota limit scenario.)

If a resource group name is passed, but not a location, then prompt for location.


cptera:
I think the other pull request already has this in it?

bill-dall:
@cptera yeah - my bad. I built the other one off the wrong branch. If you approve this one then I'll commit in-order and I think the merge log looks better then.

PR #125, opened 2018-04-06
Enable linux graphics Agent

cptera:


remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -20,9 +20,10 @@
 # THE SOFTWARE.
 #
 
-if [[ $# -eq 1 ]]
```
fwang-teradici: 1) using -lt is better 2)how about only 1 parameter passing, then variable BINARY_LOCATION is null or empty which might cause problem

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -20,9 +20,10 @@
...
+if [[ $# -eq 1 ]] || [[ $# -eq 2 ]]
 then
```
fwang-teradici: using -lt is better 2)how about only 1 parameter passing, then variable BINARY_LOCATION is null or empty which might cause problem

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -20,9 +20,10 @@
 # THE SOFTWARE.
 #
 
-if [[ $# -eq 1 ]]
+if [[ $# -eq 1 ]] || [[ $# -eq 2 ]]
```
fwang-teradici: using -lt is better 2)how about only 1 parameter passing, then variable BINARY_LOCATION is null or empty which might cause problem

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -53,64 +54,80 @@ else
...
-  sudo yum -y update
-  sudo yum -y install kernel-devel
-  sudo rpm -Uvh --quiet https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-  sudo yum -y install dkms
+    echo "-->Updating" | tee -a "$INST_LOG_FILE"
+    sudo yum -y update
+    echo "-->Installing kernel-devel and dkms" | tee -a "$INST_LOG_FILE"
+    sudo yum -y install kernel-devel
+    sudo rpm -Uvh --quiet https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+    sudo yum -y install dkms
```
bdyck: I feel like we should check for success/failure for these package installations, since there isn't really much point in continuing if they fail...

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -53,64 +54,80 @@ else
...
     fi
 
     return $exitCode
 }
 
 install_nvidia_driver()
 {
-    local FILE_NAME="NVIDIA-Linux-x86_64-384.73-grid.run"
+    local FILE_NAME="NVIDIA-Linux-x86_64-384.111-grid.run"
+    echo "-->Downloading nVidiai Driver ${FILE_NAME} from ${BINARY_LOCATION}" | tee -a "$INST_LOG_FILE"
```
bdyck: This should be "nVidia"

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -397,13 +416,13 @@ register_pcoip_license()
...
     # Install and setup the Sumo Collector
-    echo "-->Install SumoLogic collector"
+    echo "-->Install SumoLogic collector" | tee -a "$INST_LOG_FILE"
     mkdir /tmp/sumo
     wget "https://collectors.sumologic.com/rest/download/linux/64" -O /tmp/sumo/SumoCollector.sh && sudo chmod +x /tmp/sumo/SumoCollector.sh
     wget "$STORAGEURI/user.properties$SAS_TOKEN" -O /tmp/sumo/user.properties
     wget "$STORAGEURI/sumo-agent-vm-linux.json$SAS_TOKEN" -O /tmp/sumo/sumo-agent-vm-linux.json
     JSON_FILE=/tmp/sumo/sumo-agent-vm-linux.json
-    echo "Attemtping to set sumo collector ID to: " "$COLLECTOR_ID"
+    echo "-->Attempting to set sumo collector ID to: ${COLLECTOR_ID}" | tee -a "$INST_LOG_FILE"
```
bdyck: I'm not sure if we want to log the collector ID itself...

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -450,19 +470,19 @@ AGENT_TYPE=$(tr '[:upper:]' '[:lower:]' <<<"$AGENT_TYPE")
...
-    echo "unknown agent type $AGENT_TYPE."
+    echo "Unknown agent type $AGENT_TYPE." | tee -a "$INST_LOG_FILE"
     # let's define exit code 105 for this case
     exit 105
 fi
 
 if [[ -f "$INST_LOG_FILE" ]]
 then
-    INST_LAST_STEP=$(tail -1 "$INST_LOG_FILE") 
+    INST_LAST_STEP=$(cat "$INST_LOG_FILE" | grep "^step*" | tail -1)
```
bdyck: This should be changed to `grep "^step*" "${INST_LOG_FILE}" | tail -1`.

remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
bdyck: This seems like a potential bug down the road... I think the script should probably check that `BINARY_LOCATION` is set correctly when it tries to use it on successive invocations.

PR #126, opened 2018-04-18
Fix linux join domain

fwang-teradici:


remote-workstations/new-agent-vm/Install-PCoIPAgent.sh
```diff
@@ -391,22 +418,47 @@ install_pcoip_agent()
...
     # Install the PCoIP Agent
-    echo "-->Install the PCoIP $AGENT_TYPE agent" | tee -a "$INST_LOG_FILE"
+    local dispName=("first" "second" "third")
     for idx in {1..3}
     do
+        echo "-->Install the PCoIP $AGENT_TYPE agent at ${dispName[idx -1]} time" | tee -a "$INST_LOG_FILE"
         sudo yum -y install "pcoip-agent-$AGENT_TYPE"
         exitCode=$?
 
+        if [[ $exitCode -eq 0 ]]
```
bill-dall: So the agent installer does not set firewall?

fwang-teradici: it does set, and it works before. But now the firewalld service itself is stopped at this stage which cause installer set firewall failed. (I do not know why). So restart firewall service and set firewall after installation is done which solve the issue. This issue is occurred since last week

bill-dall: Okay. Cool. Maybe add some comment about 'pcoip agent installer usually does this but it can fail if the firewall service is disabled, so re-doing here.' since otherwise it would looks strange and duplicated to a future reviewer.

cptera: would enabling the firewall before running install make this easier?

fwang-teradici: At the very beginning the firewall is running. so do check and enable. I will at comment like bill suggestion

PR #127, opened 2018-04-20
Multi region

bill-dall:
Multi-region deployments and a number of misc. fixes and prompting improvements.


bill-dall:
Note that from an automation perspective we should be able to get exactly the same behaviour as today (and others!) but the automation will have to provide more command-line parameters than before. That felt better to me than having a bunch of defaults that were unclear why (Why? Because automation wanted it that way. Then it should just tell us...)

Deploy-CAM.ps1
```diff
@@ -1484,7 +1484,7 @@ function New-ConnectionServiceDeployment() {
             
             Write-Host "Checking available resource group for connection service number $connectionServiceNumber"
 
-            $csRGName = $RGName + "-CS" + $connectionServiceNumber
+            $csRGName = $RGName + "-CN" + $connectionServiceNumber
```
cptera: Note for whoever does the story: this is going to break automation

Deploy-CAM.ps1
```diff
@@ -3170,32 +3258,192 @@ if(-not $keyVaultProviderExists) {
...
+            break
+        }
+    }
+}
+
+# Now we (might) have a $ResourceGroupName that (might) point to a resource group that has a CAM keyvault in it.
+# Let's double check what type.
+# At this point we have a subscription and a root resource group - check if there is already a deployment in it
+try{
+    $CAMRootKeyvault = $null
```
cptera: Not a big deal, but instatiation should probably be outside the try, catch

Deploy-CAM.ps1
```diff
@@ -3170,32 +3258,192 @@ if(-not $keyVaultProviderExists) {
...
+        Write-Host "More than one CAM Key Vault found in this resource group."
+        Write-Host "Please move or remove all but one."
+        return   # early return!
+    }
+
+    Write-Host "The resource group $ResourceGroupName has a CAM deployment."
+    Write-Host "Using key vault $($CAMRootKeyvault.Name)"
+
+    # Ensure this user account can get secrets from the key vault
+    $hasKVAccess = Set-KeyVaultAccess($CAMRootKeyvault)
```
cptera: shouldn't this be 'Set-KeyVaultAccess -camRootKeyvalut $CAMRootKeyvault'?

cptera:
Looks good to me, but I think someone should try running it off of the branch before merging since a lot of stuff changed from the looks of it (or run automation against this branch or something)

Deploy-CAM.ps1
bill-dall: Good point. Fixed.

Deploy-CAM.ps1
bill-dall: Yes. I think it worked because there was only one parameter but it is fixed.

Deploy-CAM.ps1
```diff
@@ -1754,7 +1754,8 @@ function New-ConnectionServiceDeployment() {
                         -ResourceGroupName $csRGName `
```
cptera: Looks liek we're still using CS# for the deployment name. Do we want to change that to CN#?

bill-dall: Yeah I didn't rename all the internal variables since that would be an even more massive change. or the keyvault secrets since that would hose upgrades. I feel we should make a backlog story about 'completing renaming connection service to connector' to deal with those things if that makes sense to you. I think the resource group naming is properly handled though - or did I miss something?

PR #136, opened 2018-05-08
add parameter camUserGroup

fwang-teradici:


bill-dall:
Thanks. I was worried you'd added more prompts etc, but very happy to see a new command-line option there - that's a good tradeoff for today.

PR #142, opened 2018-05-28
TSW-71918: Deploy-CAM.ps1 now takes to top entry for AzureRM and Azure

rohml:
When doing the Confirm-ModuleVersion check AzureRM and Azure takes only the top entry of the object for the comparison. At this point $AzureRMModule and $AzureModule will have contents, so this should always return a result for comparison.

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -2648,7 +2648,7 @@ function Confirm-ModuleVersion()
     $AzureRMModule = Get-Module -ListAvailable -Name "AzureRM"
     if ( $AzureRMModule ) {
         # have an AzureRM version - check that.
-        if ( [version]$AzureRMModule.Version.ToString() -lt [version]$MinAzureRMVersion) {
+        if ( [version]$AzureRMModule[0].Version.ToString() -lt [version]$MinAzureRMVersion) {
             Write-Host ("AzureRM module version must be equal or greater than " + $MinAzureRMVersion)
             return $false
         }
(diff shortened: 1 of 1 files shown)


bill-dall:
Are we guaranteed the arrays are in the order we want or can/should we do a sort?

Deploy-CAM.ps1
```diff
@@ -2648,7 +2648,7 @@ function Confirm-ModuleVersion()
     $AzureRMModule = Get-Module -ListAvailable -Name "AzureRM"
     if ( $AzureRMModule ) {
         # have an AzureRM version - check that.
-        if ( [version]$AzureRMModule.Version.ToString() -lt [version]$MinAzureRMVersion) {
```
fwang-teradici: how about there is only one version. Is it still an array?

rohml: Yes, it returns an object and will always have an index 0, currently works perfectly fine with one version installed. 

All I have seen, found and tested around has had the list sorted with the newest on top. But I can't confirm this is the design, no documentation saying this is the case. I can definitely implement a sort with a couple lines, should I do this?

bill-dall: @rohml yeah in my experience if something magically works and isn't by any design then it often magically unworks at an inopportune time in the future. So if it's easy to sort, I'd suggest we do that. Thanks!

Deploy-CAM.ps1
```diff
@@ -2648,7 +2648,7 @@ function Confirm-ModuleVersion()
     $AzureRMModule = Get-Module -ListAvailable -Name "AzureRM"
     if ( $AzureRMModule ) {
         # have an AzureRM version - check that.
-        if ( [version]$AzureRMModule[0].Version.ToString() -lt [version]$MinAzureRMVersion) {
+        if (( $AzureRMModule.Version -ge [version]$MinAzureRMVersion) -eq $False) {
```
bill-dall: This checks the whole array of module versions?

rohml: Yep, we get the version list, then filter it for everything that is greater than the version number. If there is something over the version number it will be found in there, if not then it will read as false. So we check for a match with false. 
Its very similar to the previous iteration, just instead of finding the top result, it finds all the matches.
The matching to false is because when returning a version that isn't actually 'true' so I can't -not it.

bill-dall: Thanks for the demo. I didn't know that operators like '.' and '-ge' can map to 'select-object' and 'where-object' which is actually kinda' cool but makes my brain hurt.

PR #147, opened 2018-06-12
TSW-70405: Password never expires

rohml:
Set-ADUser with the username from the admin credentials to set the password to never expire for a CAM-in-the-box deploy. 

Should be able to test easily enough on a CAM-in-the-box deploy by running 
Get-ADUser -Identity <username> -Properties PasswordNeverExpires 
in the vm-dc Powershell. PasswordNeverExpires will be set to true.

### DIFF
--- a/root/configure-dc-and-ca/Install-DC-and-CA.ps1
+++ b/root/configure-dc-and-ca/Install-DC-and-CA.ps1
@@ -144,7 +144,8 @@ configuration CreateDCCA
             GetScript  = { @{ Result = "Set_Admin_Password_Permanent"}}
 
             TestScript = {
-                Test-Path -Path "C:\adminUpdate" #If file isn't found, then run SetScript.
+                # If file isn't found, then run SetScript.
+                Test-Path -Path "C:\adminUpdate" 
                 }
 
             SetScript  = {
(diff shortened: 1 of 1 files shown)


root/configure-dc-and-ca/Install-DC-and-CA.ps1
```diff
@@ -138,6 +138,44 @@ configuration CreateDCCA
...
+                $retry = 3
+                while(-not (Set-Password-Permanent -uName $($using:Admincreds.Username))){
+                    if(($retry--) -le 0){
+                        Write-Host "Failure to connect to Azure, Admin password not set to permanent."
+                    }
+                    Start-Sleep -seconds 30 # Wait for another attempt
+                }
+
+                $fileTest = New-Item "C:\adminUpdate" -type File
+                Set-Content -Path $fileTest -Value "Admin password permanent."
```
cptera: This gets set even if it fails?

rohml: New update should fix this. Should only call when PasswordNeverExpires is set to true. Re-organized a bit to correct for some bad logic handling failed connections.

root/configure-dc-and-ca/Install-DC-and-CA.ps1
```diff
@@ -138,6 +138,49 @@ configuration CreateDCCA
...
+                $retry = 3
+                $passwordStatus = $false
+                while(-not ($passwordStatus = (Set-Password-Permanent -uName $($using:Admincreds.Username))) -and ($retry-- -gt 0)){
+                    # Wait for another attempt
+                    Start-Sleep -seconds 30 
+                }
+
+                if($passwordStatus){ 
+                    # Create file if successful.
+                    $fileTest = New-Item "C:\adminUpdate" -type File
```
cptera: If this file isn't created, the DSC will fail the SetScript and run it again. Is this what you intended?

PR #148, opened 2018-06-20
TSW-71560: Added a check/register for Microsoft.Compute and Microsoft.Network

rohml:
Added a function to check for a resource that is available. Ported much of the code from the original Microsoft.Keyvault check.

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -3017,6 +3017,35 @@ function Write-Host-Warning() {
     Write-Host ("`n$message") -ForegroundColor Red
 }
 
+# Check if a resource is currently registered to the Subscription
+function checkResource(){
+    param(
+        [parameter(Mandatory=$true)]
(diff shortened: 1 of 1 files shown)


Deploy-CAM.ps1
```diff
@@ -3558,16 +3587,12 @@ $selectedSubcriptionId = $subscriptionsToDisplay[$chosenSubscriptionIndex].Subsc
 
 # Now we have the subscription set. Ensure it has keyvault resource provider
 $keyVaultProviderExists = [bool](Get-AzureRmResourceProvider | Where-Object {$_.ProviderNamespace -eq "Microsoft.Keyvault"})
```
cptera: Is this line still needed?

rohml: Nope, was from the previous code, not sure why me from a week ago didn't remove it. Updated.

PR #156, opened 2018-07-20
Update the ExternalRoutableIP at bootup time.

jackyycheng:


connection-service/new-cm-sg-ls-vm/cm_setup_0.1.sh
```diff
@@ -145,6 +145,25 @@ if [ ${sge} = "true" ]; then
 		cp -f SecurityGateway.conf /etc/SecurityGateway.conf
 
 		echo "Finished setting up SG"
+
```
fwang-teradici: how about put the code in a function and invoke the function here

connection-service/new-cm-sg-ls-vm/cm_setup_0.1.sh
```diff
@@ -23,6 +23,29 @@
...
+	# If anyone tries to shutdown the CM/SG, the public IP address of CM/SG would be different next time when you bootup. 
+	# This script will run at bootup time and change the ExternalRoutableIP 
+	echo "Setting up bootup script"
+	file_path=/root/update_ip.sh
+cat <<EOF >$file_path
+#!/bin/bash
+# Make sure SG and CM start up first
+sleep 15
+mypublicip=\$(dig +short myip.opendns.com @resolver1.opendns.com)
+cp -n /etc/SecurityGateway.conf /etc/SecurityGateway.conf.orig
```
cptera: I think the original /etc/SecurityGateway.con.orig is the default SG.conf file when you install a CM/SG (ie, a good reference config file). I'm not sure if we want to overwrite everytime on boot? On the other hand, I don't think customers can modify CM/SG settings easily so they might not need the reference file and maybe this isn't an issue

PR #175, opened 2019-02-14
Make ConnSrvr DSC more resilient:

cptera:
* DSC will loop through all available DCs for root cert and stop at the first successful one


cptera:
implified view of changes: https://github.com/teradici/deploy/pull/175/files?w=1

PR #176, opened 2019-02-22
Added support for configuring CAM Management User Group:

cptera:
* Added new KeyVault setting
* Create new KeyVault setting on Update

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -3654,21 +3654,24 @@ function Update-CAMAzureKeyVault() {
         -VaultName $VaultName `
         -Name "camManagementUserGroup" `
         -ErrorAction stop
-    if (-not $secret) {
+
+    do {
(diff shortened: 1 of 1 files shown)


Deploy-CAM.ps1
```diff
@@ -4222,6 +4269,16 @@ else {
         radiusServerPort = $radiusServerPort 
         radiusSharedSecret = $radiusSharedSecret
     }
+
+    if ($camManagementUserGroup -eq $null -and (-not $ignorePrompts)) {
+        $camManagementUserGroup = Read-Host "Enter the User Group Name or the Distinguished Name for the User Group to log into the CAM Management Interface. Default is 'Domain Admins'. (eg, 'Domain Admins' or 'CN=Domain Admins,CN=Users,DC=example,DC=com')"
```
bdyck: I think that in the case where someone wants to customise the management user group, we should require the full DN for the group so that there is no possibility of getting the wrong group if we try to figure out the DN later (as we currently do for the "Domain Admins" group).

PR #178, opened 2019-05-06
using new pcoip agent distribute format and new 64 bit installer

fwang-teradici:


remote-workstations/new-agent-vm/Install-PCoIPAgent.ps1
```diff
@@ -282,32 +290,55 @@ Configuration InstallPCoIPAgent
...
+                    }
+                }
+
+                # download installer
+                $meta=(Get-Content -Raw -Path  $destFile | ConvertFrom-Json)
+                $installerFileName =$meta.filename
+
+                $downloadUrl =  $pcoipAgentInstallerUrl + $installerFileName
+                
+                $destFile = $using:agentInstallerDLDirectory + '\' + $installerFileName
```
cptera: Not a big deal, but you might want to use "Join-Path" for stuff like this

remote-workstations/new-agent-vm/Install-PCoIPAgent.ps1
```diff
@@ -351,8 +382,13 @@ Configuration InstallPCoIPAgent
             GetScript  = { return 'registration'}
             
             TestScript = { 
-                cd "C:\Program Files (x86)\Teradici\PCoIP Agent"
-                 $ret = & .\pcoip-validate-license.ps1
+                $path = "C:\Program Files (x86)\Teradici\PCoIP Agent"
```
cptera: Why not use the RegKey path like in the Test-Scipt instead of doing a mix of RegPath and InstallPath checks to pick the install path

PR #188, opened 2019-06-27
set retrieveAgentState default to false

fwang-teradici:


Deploy-CAM.ps1
```diff
@@ -117,7 +117,7 @@ Param(
     $camManagementUserGroup = $null,
 
     [switch]$updateSPCredential,
-    [bool]$retrieveAgentState = $true,
+    [bool]$retrieveAgentState = $false,
     [bool]$showAgentState = $true
```
cptera: should just make this false too even though it doesn't do anything

fwang-teradici: I think true is better as default so custom only need to set retrieveAgentState is true and it shows the state (normally people like to see if the state is there)

cptera: i think we want to be consistant between v1 and v2

cptera: nevermind, looks liek thats what v2 does

PR #204, opened 2020-01-23
SecureGateways deprecated:

cptera:
* Azure has deprecated SecureGateways and role no longer exists

### DIFF
--- a/Deploy-CAM.ps1
+++ b/Deploy-CAM.ps1
@@ -3145,8 +3145,6 @@ function Get-CAMRoleDefinition() {
             # Remove ability to access route filters and tables
             'Microsoft.Network/routeFilters/*'
             'Microsoft.Network/routeTables/*'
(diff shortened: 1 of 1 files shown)


jackyycheng:
Actually, Can we have try and catch block? So in the future it would not fail the installation?

Deploy-CAM.ps1
```diff
@@ -3145,8 +3145,6 @@ function Get-CAMRoleDefinition() {
             # Remove ability to access route filters and tables
             'Microsoft.Network/routeFilters/*'
             'Microsoft.Network/routeTables/*'
-            # Remove ability to access secure gateways
```
jackyycheng: Let's have try and catch on removing the roles.

cptera: try/catch might get messy since it's a batch thing and we want it to error out if it fails, but i can verify that the role exists if the same thing happens again

PR #208, opened 2020-03-17
Update connection server VM:

cptera:
* Use the latest version


cptera:
it seems to be working okay now, MS may have resolved the issue we were previously seeing themselves

